PR #2, opened 2020-07-23
Add dos reload site

marcosholgado:
Description:
This PR adds a new security site with an infinite loop calling location.reload -> https://app.asana.com/0/1125189844152671/1184979058928819


kdzwinel:
I added title and a home button just to keep the convention. LGTM, thank you!

PR #4, opened 2020-08-10
tidy up, surrogate and links pages

brindy:
Adds 
* A page with a surrogate
* A page with a link that opens in a new window
* A page with a link that tries to close itself

### DIFF
--- a/index.html
+++ b/index.html
@@ -7,29 +7,32 @@
 </head>
 <body>
 
-<h1>Client Test Pages</h1>
+<h1>Private Test Pages</h1>
 
 <p>This site contains a collection of pages with known web content (e.g. trackers, web functionality, etc) that can be used for testing by the extensions and mobile apps.<p>
 
 <h2>Trackers</h2>
 
 <ul>
-	<li><a href="./trackers/1major-via-script.html">1 major tracker loaded via script</a>
-	<li><a href="./trackers/1major-via-img.html">1 major tracker loaded via img</a>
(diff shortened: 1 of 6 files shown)


index.html
```diff
@@ -7,29 +7,31 @@
...
+	<li><a href="./trackers/document-fragment.html">Image loaded via document fragment</a></li>
 </ul>
 
 <h2>Browser Features</h2>
 
 <ul>
-	<li><a href="./features/geolocation.html">Geolocation</a>
+	<li><a href="./features/geolocation.html">Geolocation</a></li>
+	<li><a href="./features/target-blank.html">Links Open in New Window</a>
+    <li><a href="./features/closeable.html">window.close() Link</a>
```
kdzwinel, line 28: 👀 I wish there was one more space there…

trackers/1major-with-surrrogate.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width">
+  <title>1 major tracker loaded via script</title>
+
+  <script src="//static.doubleclick.net/tracker.js"></script>
```
kdzwinel, line 8: I don't think we surrogate this url? You should do `doubleclick.net/tag/js/gpt.js` or `doubleclick.net/instream/ad_status.js` IMO.

Also, `<title>` needs an update.

index.html
brindy: this is why the internet is running out of space!

PR #6, opened 2020-08-26
Add fetch test page

SlayterDev:
Test page which loads a tracker using the fetch api


kdzwinel:
That request seems to get blocked by CORS even if it's not blocked by the tracker blocker?

SlayterDev:
Hmmm looked like it worked in my testing. We can switch the tracker if needed?

kdzwinel (approved):
Thanks for a quick fix!

PR #12, opened 2020-11-27
Privacy Protection Test Pages: Fingerprinting 👆

kdzwinel:
Tech design: https://app.asana.com/0/0/1198919147577399/f
Live: https://privacy-test-pages.glitch.me/privacy-protections/fingerprinting/

This PR adds fingerprinting test page that collects over 100 different data points and allows to store and compare them.


privacy-protections/request-blocking/main.js
```diff
@@ -442,6 +442,7 @@ function runTests() {
     startButton.setAttribute('disabled', 'disabled');
     downloadButton.removeAttribute('disabled');
     results.results.length = 0;
+    results.date = (new Date()).toUTCString();
```
kdzwinel, line 445: small addition to the previous test - storing current date with output data

server.js
```diff
@@ -42,4 +51,9 @@ app.get('/block-me/server-sent-events', (req, res) => {
 // dummy CSP report endopoint
 app.post('/block-me/csp', (req, res) => {
     return res.sendStatus(200);
+});
+
+// reflects request headers back
+app.get('/reflect-headers', (req, res) => {
+    return res.json({headers: req.headers});
```
kdzwinel, line 58: that's how we get info about headers sent - this endpoint returns them as a JSON file

privacy-protections/fingerprinting/helpers/constants.js
```diff
@@ -0,0 +1,9 @@
+// fingerprint.js + various other sources
```
kdzwinel, line 1: I've put all the boring and long stuff in this file to keep others a bit more redable.

privacy-protections/fingerprinting/helpers/diff.js
```diff
@@ -0,0 +1,3 @@
+// https://github.com/flitbit/diff
```
kdzwinel, line 1: Import for deep diffing objects

privacy-protections/fingerprinting/helpers/tests.js
```diff
@@ -0,0 +1,1292 @@
...
+    {
+        id: 'navigator.vendorSub',
+        category: 'navigator',
+        getValue: () => navigator.vendorSub
+    },
+    {
+        id: 'navigator.mimeTypes',
+        category: 'navigator',
+        getValue: () => {
+            const results = {};
```
kdzwinel, line 118: in many cases I, somehow forcefully, create objects rather than arrays (even if arrays would make more sense). The reason is diffing - if one element in the middle of the array is missing the diff will be huge because all other items are shifted, for objects this is not a problem.

privacy-protections/fingerprinting/helpers/tests.js
```diff
@@ -0,0 +1,1292 @@
...
+                    force: event.touches[0].force
+                });
+            });
+
+            return promise;
+        }
+    },
+    {
+        id: 'ondevicelight',
+        category: 'events',
```
kdzwinel, line 1278: I have not included 'resize' here as the resize event itself doesn't seem to have any additional data (sth like `event.change`). So in order to test resize one can just resize the window and run test again. LMK if I'm missing something here.

privacy-protections/request-blocking/index.html
```diff
@@ -35,6 +35,13 @@
             </li>
         </ul>
 
+        <p><strong>Legend:</strong></p>
```
kdzwinel, line 38: Something that we discussed and I forgot to add in the previous PR - a legend for the request blocking page.

kdzwinel (commented):
@Charlie-belmer  LMK what you think!

privacy-protections/fingerprinting/helpers/tests.js
```diff
@@ -0,0 +1,1292 @@
...
+                }
+
+                result[propName] = navigator[propName];
+            });
+
+            return result;
+        }
+    },
+    {
+        id: 'screen',
```
Charlie-belmer, line 1176: Is this something we should consolidate into a stand alone function? Seems we could pass an object name and run through this, so we could more easily add additional objects to enumerate in the future if we wanted to beyond just screen and navigator.

privacy-protections/fingerprinting/main.js
```diff
@@ -0,0 +1,230 @@
...
+
+      compareResults(oldResults, results);
+    };
+    reader.readAsText(compareWithFile.files[0]);
+})
+
+downloadButton.addEventListener('click', () => downloadTheResults());
+
+// run tests if button was clicked or…
+startButton.addEventListener('click', () => runTests());
```
Charlie-belmer, line 225: One thing I thought of is that by making this run on click, we won't be testing any race conditions around fingerprinting. (For instance, a site fingerprints with a script in head right away, without waiting for the dom to load or fully load). Depending on how defenses are implemented, they may or may not defend against that. Probably not a problem for our use case, but maybe something to be aware of.

Charlie-belmer (approved):
Looks good to me!

privacy-protections/fingerprinting/main.js
kdzwinel: Good point, I haven't thought about this. However(!) there is a way to run test on page load - see code below. I added it mostly for automated tests, but as far as I understand it would also be useful in the scenario you are describing.

privacy-protections/fingerprinting/helpers/tests.js
kdzwinel: Ah, really good tip, that's clearly almost identical code. I extracted it to a function AND while playing with it found out how to list even more properties (my use of `Object.keys` was wrong). I also decided to include properties with complex values but just replace the values with simple placeholders ('~object~').

moar

kdzwinel:
Thanks man! Good stuff. I addressed all of your comments and added a bit of extra magic.

PR #13, opened 2020-12-10
Storage Blocking Page

kdzwinel:
Tech Design: https://app.asana.com/0/0/1198919147577403/f
Live Page: https://privacy-test-pages.glitch.me/privacy-protections/storage-blocking/

Two notes:
- I messed up rebase, please don't look at the list of commits 🙈
- I included some small changes to request-blocking (related to introducing our new faux tracker bad.third-party.site)


privacy-protections/request-blocking/service-worker.js
```diff
@@ -2,9 +2,13 @@ self.addEventListener('install', (evt) => {
     self.skipWaiting();
 });
 
+self.addEventListener('activate', (evt) => {
+    evt.waitUntil(clients.claim());
```
kdzwinel, line 6: this makes the new service worker start proxying requests right away (and take it over from previous SW that may exist).

privacy-protections/request-blocking/service-worker.js
```diff
@@ -2,9 +2,13 @@ self.addEventListener('install', (evt) => {
...
+});
+
 self.addEventListener('message', (event) => {
-    if (event.data === 'fetch') {
-        fetch(`./block-me/fetch.json?${Math.random()}`)
+    if (event.data.action && event.data.action === 'fetch') {
```
kdzwinel, line 10: ditto not hardcoding url

privacy-protections/request-blocking/worker.js
```diff
@@ -1,8 +1,12 @@
...
-    .then(r => r.json())
-    .then(data => {
-        if (data.data.includes('fetch loaded')) {
-            postMessage('worker fetch loaded 👍');
-        }
-    })
-    .catch(e => postMessage('worker fetch failed'));
\ No newline at end of file
+self.addEventListener('message', msg => {
+    if (msg.data.action && msg.data.action === 'fetch') {
```
kdzwinel, line 2: ditto

privacy-protections/storage-blocking/helpers/idb-wrapper.js
```diff
@@ -0,0 +1,72 @@
+// https://raw.githubusercontent.com/msandrini/minimal-indexed-db/master/index.js
```
kdzwinel, line 1: just a indexeddb wrapper. I thought I can use native for the simple thing I'm doing, but indexeddb API is so bad that I gave up.

privacy-protections/storage-blocking/iframe.js
```diff
@@ -0,0 +1,168 @@
+const tests = [
```
kdzwinel, line 1: This is a subgroup of tests in main.js

privacy-protections/storage-blocking/main.js
```diff
@@ -0,0 +1,411 @@
+const THIRD_PARTY_ORIGIN = 'https://good.third-party.site';
```
kdzwinel, line 1: third party that we are using for a 3p cookie test and 3p storage via iframe test (this is the domain that should *not* be blocked)

privacy-protections/storage-blocking/main.js
```diff
@@ -0,0 +1,411 @@
...
+            return caches.open('data').then((cache) => {
+                return cache.match('/cache-api-response')
+                    .then(r => r.text());
+            });
+        }
+    },
+    {
+        id: 'browser cache',
+        store: (data) => {
+            // already done before all tests
```
kdzwinel, line 200: see `storeData()`. We can't really pass a number to server (e.g. `/cached-random-number?number=${data}`) because we won't be able to guess that url later on. We have to have one url (`/cached-random-number`) that we will query to warm up the cache and retrieve the value. And for that we have to do this operation first and use the number server produces to fill all other storage.

server.js
```diff
@@ -11,11 +11,12 @@ const listener = app.listen(port, () => {
 // serve all static files
 app.use(express.static('.', {
     setHeaders: (res, path) => {
-        res.set("Access-Control-Allow-Origin", "*");
+        res.set('Access-Control-Allow-Origin', '*');
+        res.set('Timing-Allow-Origin', '*');
```
kdzwinel, line 15: this allows 3p websites to access timing measures for that requests (how long did it load etc.)- we use that for request blocking

server.js
```diff
@@ -11,11 +11,12 @@ const listener = app.listen(port, () => {
 // serve all static files
 app.use(express.static('.', {
     setHeaders: (res, path) => {
-        res.set("Access-Control-Allow-Origin", "*");
+        res.set('Access-Control-Allow-Origin', '*');
```
kdzwinel, line 14: this allows 3p websites to access content of requests (we use that for request blocking)

server.js
```diff
@@ -50,10 +53,37 @@ app.get('/block-me/server-sent-events', (req, res) => {
...
+    res.set('Access-Control-Allow-Credentials', 'true');
+    res.set('Timing-Allow-Origin', '*');
+
     return res.json({headers: req.headers});
+});
+
+// sets a cookie with provided value
+app.get('/set-cookie', (req, res) => {
+    res.set('Access-Control-Allow-Origin', req.headers.origin || '*');
+    res.set('Access-Control-Allow-Credentials', 'true');
```
kdzwinel, line 72: we need to set both of those of that for the cookie to be actually set nowadays. And we can't use '*' for ACAO. TIL!

kdzwinel (commented):
@Charlie-belmer PTAL when you get a chance :)

Charlie-belmer (commented):
Everything looks good to me! The only thing I saw was the subset of tests in iframe that duplicate the main one. It's not the cleanest solution, but what if we did something like had a file commonTests.js that we imported into both the main and iframe, then specific tests could append to the tests array?

Mainly, I'm thinking about if we come by later and update a test in one area, but forget to update the other.

kdzwinel:
Good thinking! I extracted them to a separate file as you suggest.

PR #17, opened 2020-12-17
Add endpoint for auto update via GH hooks.

kdzwinel:
This creates an endpoint on the server that fetches new version of the app whenever github lets us know about changes (via a webhook). We are protecting form the abuse by using a simple password that's not stored anywhere in the code.

This follows the https://support.glitch.com/t/tutorial-how-to-auto-update-your-project-with-github/8124 tutorial.

### DIFF
--- a/server.js
+++ b/server.js
@@ -42,7 +42,7 @@ app.post('/git', (req, res) => {
     const hmac = crypto.createHmac('sha1', process.env.SECRET);
     const sig  = 'sha1=' + hmac.update(JSON.stringify(req.body)).digest('hex');
 
-    if (req.headers['x-github-event'] == 'push' && crypto.timingSafeEqual(sig, req.headers['x-hub-signature'])) { 
+    if (req.headers['x-github-event'] === 'push' && crypto.timingSafeEqual(sig, req.headers['x-hub-signature'])) { 
         cmd.run('chmod 777 git.sh'); /* :/ Fix no perms after updating */
         cmd.get('./git.sh', (err, data) => {  // Run our script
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -35,14 +37,26 @@ app.use(express.static('.', {
...
-//     }
-
-//     return res.sendStatus(200);
-// });
+// endpoint for updating the app (https://support.glitch.com/t/tutorial-how-to-auto-update-your-project-with-github/8124)
+app.post('/git', (req, res) => {
+    const hmac = crypto.createHmac('sha1', process.env.SECRET);
+    const sig  = 'sha1=' + hmac.update(JSON.stringify(req.body)).digest('hex');
+
+    if (req.headers['x-github-event'] == 'push' && sig == req.headers['x-hub-signature']) {
```
Charlie-belmer, line 45: It doesn't matter for this endpoint currently, but the '==' string comparison is vulnerable to a timing attack, so it's possible for someone to craft a body and find the right signature that would pass without too much effort. To prevent this, we can use [crypto.timingSafeEqual](https://nodejs.org/api/crypto.html#crypto_crypto_timingsafeequal_a_b) instead of == for the signature comparison.

That way if we ever do anything sensitive here, it won't be vulnerable.

Charlie-belmer (commented):
Looks pretty good! One minor thing on hash checking which isn't that important as written, but probably good to do in case one day this becomes more sensitive.

server.js
kdzwinel: TIL! I had no idea. Thank you for bringing this up.

PR #18, opened 2021-01-15
Storage test site - correctly set a 3p cookie from an iframe.

kdzwinel:
We combined some code during PR review that we shouldn't have. Cookie for 1p frame should be set with a different sameSite value than cookie for the 3p frame. Right now this cookie doesn't get set at all (silently) and makes all browsers show that it was blocked.


SlayterDev (approved):
LGTM 👍

PR #19, opened 2021-01-26
Privacy test pages - small fixes

kdzwinel:
Storage test page was not retrieving data from all the places we were saving the data to.
HTTPS test page was not reporting correctly subrequest updates because server may not know if connection was http or https if it's behind a proxy (and that's usually the case).


kdzwinel:
I'm seamlessly self-merging because those fixes are small.

PR #20, opened 2021-01-26
Request Blocking - improve detection of loaded request

kdzwinel:
Decide if request has loaded or not by looking at the server-timing data instead of duration and protocol (which were a bit unreliable in some browsers).


kdzwinel:
Another shamelessly self-merged quick fix.

IdkSaveme (commented):
2

PR #22, opened 2021-02-05
Add crawler detection page

sammacbeth:
Basic test page to allow tracker-radar-collector to test if it can be detected as a crawler.

### DIFF
--- a/crawler/index.html
+++ b/crawler/index.html
@@ -6,9 +6,11 @@
 </head>  
 
 <body>
-<p><a href="../index.html">[Home]</a></p>
+<p><a href="../index.html">[Home]</a> ↣ <strong>[Crawler Detectability test]</strong></p>
 
 <p>This page tests against common ways of detecting headless browsers.</p>
+
+<p>A pass means that this looks like a normal Chrome browser, fail means it's Headless Chrome.</p>
 
 <table>
     <tr>
--- a/crawler/index.html
+++ b/crawler/index.html
@@ -1,49 +1,54 @@
 <!DOCTYPE html>
-<html>
-  <head>
-    <meta encoding='utf-8'/>
-    <title>Headless detector</title>
-  </head>
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width">
+  <title>Crawler Detectability test</title>
+</head>  
 
-  <body>
-    <table>
-        <tr>
-            <th>Test</th>
-            <th>Result</th>
-        </tr>
-        <tbody id="results"></tbody>
-    </table>
-    <template id="result-row">
-        <tr>
-            <td class="key"></td>
-            <td class="value"></td>
-        </tr>
-    </template>
-    <script>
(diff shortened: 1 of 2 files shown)


crawler/index.html
```diff
@@ -0,0 +1,49 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta encoding='utf-8'/>
+    <title>Headless detector</title>
+  </head>
+
+  <body>
```
kdzwinel, line 8: Can you please add breadcrumbs and a description like we have on other pages?

header

Sth like "This page collects various tests for detecting headless browsers."

crawler/index.html
```diff
@@ -0,0 +1,49 @@
...
+                if (result) {
+                    fetch(`./fail.json?test=${test}&${detectValue}`);
+                }
+            }
+
+            displayResult('userAgent', /HeadlessChrome/.test(window.navigator.userAgent), window.navigator.userAgent);
+            displayResult('plugins', navigator.plugins.length == 0, navigator.plugins.length);
+            displayResult('languages', navigator.languages == "", navigator.languages);
+            displayResult('webdriver', !!navigator.webdriver, navigator.webdriver)
+            displayResult('window.chrome', !window.chrome || !window.chrome.runtime, JSON.stringify(window.chrome))
```
kdzwinel, line 41: FYI there is no `window.chrome.runtime` in Chrome Canary (90) and my plugins are in fact empty. Probably not a big deal, but shows that those tests are somehow not reliable.

crawler/index.html
```diff
@@ -0,0 +1,49 @@
...
+            const ul = document.getElementById('results');
+            const template = document.getElementById('result-row');
+            
+            function displayResult(test, result, detectValue) {
+                const listEntry = template.content.cloneNode(true);
+                listEntry.querySelector('.key').textContent = test;
+                listEntry.querySelector('.value').textContent = result ? 'fail' : 'pass';
+                ul.appendChild(listEntry)
+                if (result) {
+                    fetch(`./fail.json?test=${test}&${detectValue}`);
```
kdzwinel, line 33: I like this :) I'd add a note that this is used for easy access to test results from the crawler's output.

kdzwinel (changes requested):
Very neat! Could you please add a link to this page from https://github.com/duckduckgo/privacy-test-pages/blob/gh-pages/index.html ? We probably need "Other" category because it doesn't seem to fit anywhere else.

crawler/index.html
sammacbeth: The articles around this also stress that some of these methods only work with certain versions. Having this also as a webpage means we can also test against normal chrome, and remove tests which no longer pass for real browsers.

sammacbeth:
Thanks @kdzwinel - sorry I missed the docs on making a new page. I've updated the page to match the template and address your other comments.

kdzwinel (approved):
No worries and thanks again for adding this page!

PR #23, opened 2021-02-07
Add more canvas fingerprinting tests

jonathanKingston:


privacy-protections/fingerprinting/helpers/tests.js
```diff
@@ -1018,62 +1034,66 @@ const tests = [
...
+        id: 'canvas-2d-offscreen-todataurl',
+        category: 'full-fingerprints',
+        getValue: () => {
+            const offscreen = new OffscreenCanvas(2000, 200)
+
+            applyFpExampleDataToCanvas(offscreen)
+
+            // As we can't get the result directly from the offscreen canvas
+            // we then push it through to an on-screen canvas
+            const imageBitmap = offscreen.transferToImageBitmap()
```
kdzwinel, line 1087: Oh interesting, we can't make a fingerprint out of this imageBitmap directly in any way?

privacy-protections/fingerprinting/helpers/tests.js
```diff
@@ -1018,62 +1034,66 @@ const tests = [
...
+            // Verify that once the method is used that it looks the same once rendered to the screen
+            const canvasCopy = document.createElement('canvas')
+            canvasCopy.width = 2000
+            canvasCopy.height = 200
+            canvasCopy.style.display = 'inline'
+            const ctxCopy = canvasCopy.getContext('2d')
+            ctxCopy.putImageData(imageData, 0, 0)
+            addCanvasToPage(canvasCopy, 'Image data check')
+
+            return sha256(JSON.stringify([...imageData.data]));
```
kdzwinel, line 1074: We don't really need sha256 on this data, we are just using it to keep the output short, right?

privacy-protections/fingerprinting/helpers/tests.js
```diff
@@ -1018,62 +1034,66 @@ const tests = [
...
+            const imageData = ctx.getImageData(0, 0, canvas.width, canvas.height)
+
+            // Verify that once the method is used that it looks the same once rendered to the screen
+            const canvasCopy = document.createElement('canvas')
+            canvasCopy.width = 2000
+            canvasCopy.height = 200
+            canvasCopy.style.display = 'inline'
+            const ctxCopy = canvasCopy.getContext('2d')
+            ctxCopy.putImageData(imageData, 0, 0)
+            addCanvasToPage(canvasCopy, 'Image data check')
```
kdzwinel, line 1072: should we have one more of those for toDataURL?

kdzwinel (commented):
The inconsistent usage of semicolons triggers my OCD :D That's my fault though - I need to get some linting going.

Everything LGTM, but I left some questions.

privacy-protections/fingerprinting/helpers/tests.js
jonathanKingston: ImageBitmap doesn't have any API methods currently. I think it's intentional. The only other thing that can be done is used into a `new Image(...)` which to get the pixels we need another canvas (or some novel CSS shadow technique etc).

privacy-protections/fingerprinting/helpers/tests.js
jonathanKingston: Yeah we don't need it... but an entire page of 1's and 0's was pretty useless.
We could potentially offer a copy to clipboard perhaps. This just seemed the most important value for testing so we can just check that the values are different.

privacy-protections/fingerprinting/helpers/tests.js
jonathanKingston: I've added this as an image, I was mostly trying to avoid then having to take the data from an image then have to turn it into a canvas as it wouldn't potentially show what was the issue. Adding as an image instead of a canvas I think makes the most sense.

PR #24, opened 2021-02-09
Add storage test for cookies set from a 3rd party script.

sammacbeth:
Adds a new storage test for when `document.cookie` is set from a script which was loaded from a 3rd party origin.

### DIFF
--- a/privacy-protections/storage-blocking/index.html
+++ b/privacy-protections/storage-blocking/index.html
@@ -7,8 +7,8 @@
 
     <script src='./helpers/idb-wrapper.js' defer></script>
     <script src='./helpers/commonTests.js'></script>
-    <script src='./main.js' defer></script>
     <script src='https://good.third-party.site/privacy-protections/storage-blocking/3rdparty.js' defer></script>
+    <script src='./main.js' defer></script>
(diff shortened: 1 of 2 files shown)


privacy-protections/storage-blocking/index.html
```diff
@@ -8,6 +8,7 @@
     <script src='./helpers/idb-wrapper.js' defer></script>
     <script src='./helpers/commonTests.js'></script>
     <script src='./main.js' defer></script>
+    <script src='https://good.third-party.site/privacy-protections/storage-blocking/3rdparty.js' defer></script>
```
kdzwinel, line 11: If you'll run this test page with `?store` or `?retrieve` main.js will run the tests immediately and your test might be injected too late to run. How about this:

```html
<script>const tests = []</script>
<script src="…3rdparty.js" defer></script>
<script src="…main.js" defer></script>
```

kdzwinel (commented):
Nicely done! I left one comment, feel free to merge once addressed.

privacy-protections/storage-blocking/index.html
sammacbeth: Sure, I didn't know about that option, I'll make the change.

PR #25, opened 2021-02-11
Adding in a canvas drawing checker

jonathanKingston:
image

@Charlie-belmer FYI also: https://github.com/duckduckgo/privacy-test-pages/pull/23

### DIFF
--- a/features/canvas-draw.html
+++ b/features/canvas-draw.html
@@ -1,0 +1,264 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width">
+  <title>Canvas draw</title>
+</head>
+<body>
+<p><a href="../index.html">[Home]</a></p>
+
+  <p>For testing fingerprinting, ensure that both the input drawing looks the same as the output, also ensure that the input and output hashes don't match</p>
+
+  <main>
+  	<canvas id="canvas" width=300 height=300>
+   	</canvas>
+  	<canvas id="output" width=300 height=300>
+   	</canvas>
+    <div id="stats">
+    </div>
+  </main>
+
+  <aside>
+    <button id="draw-same">Draw pattern</button>
+    <p>Raw mathod calls, paste from another draw to replicate:</p>
+    <textarea id="output-data">
+    </textarea>
+  </aside>
+
+  <style>
+      #canvas, #output {
+          width: 300px;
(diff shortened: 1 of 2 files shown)


jonathanKingston:
The code picks a random colour to ensure that we check different channels etc.

features/canvas-draw.html
```diff
@@ -0,0 +1,119 @@
...
+                        this.context.beginPath();
+                        this.context.arc(x, y, radius, start, end);
+                        this.context.fill();
+                        this.context.beginPath();
+                        this.context.moveTo(x, y);
+                   	}
+                  break;
+                case 'mouseup':
+      	            this.dragging = false;
+      	            this.context.beginPath();
```
kdzwinel, line 103: too many spaces!

features/canvas-draw.html
```diff
@@ -0,0 +1,119 @@
...
+                    this.replicate()
```
kdzwinel, line 104: no semicolon! maybe we should merge that inline-eslint first :)

kdzwinel (commented):
So cool I have bunch of feature requests :)

1. can we put both DataUrl's one after another and add some `&nbsp;`s to align them so that those hashes are easy to compare (same for ImageData)
1. can we add an indicator when hashes match/don't match (can be just text saying that)
1. can we add a button that always draws the same thing so that we can test if hashes change between domains / reloads ?
1. for integration tests - can we output results to some global variable (e.g. `results`) and have a way of triggering draw (previous point) automatically (e.g. `?run` in url).

jonathanKingston:
>  can we put both DataUrl's one after another and add some &nbsp;s to align them so that those hashes are easy to compare (same for ImageData)

Done, used elements.

> can we add an indicator when hashes match/don't match (can be just text saying that)

Done

> for integration tests - can we output results to some global variable (e.g. results) and have a way of triggering draw (previous point) automatically (e.g. ?run in url).

I added a text box for now, unfortunately the image data is huge and I'm not sure if anything else is meaningful. I guess I could use toDataURL and inject it into an image perhaps.

> can we add a button that always draws the same thing so that we can test if hashes change between domains / reloads ?

Will add tomorrow :)

jonathanKingston:
@kdzwinel ready for your review again.... the run argument will provide default happiness to your day.

image

kdzwinel:
thumbs-up

Perfection ✨

PR #30, opened 2021-02-23
Adding audio fingerprinting

jonathanKingston:


kdzwinel:
The code LGTM (disclaimer: I have no idea what's going on in that audio code), but some of those new tests fail in Safari:

safari-stable

If it's a known limitation of webkit, please feel free to merge 🎶

jonathanKingston:
@kdzwinel thanks yeah neither are supported it seems so it's not an issue I think.

PR #31, opened 2021-02-23
Use a tracker, non-blocked domain for setting test cookie

sammacbeth:
~~Note this still depends on the pending DNS record for `broken.third-party.site`.~~


kdzwinel (commented):
Thanks! I think it'd be useful to keep both good.third-part.site and broken.third-party.site scenario. Now that we have broken.third-party.site I have also used it to test the iframe scenario. Finally, I added small section to the readme to explain all those test domains.

I'm reflecting this PR back to you @sammacbeth 🏓

sammacbeth (commented):
Looks great, thanks for adding these!

PR #32, opened 2021-03-04
Surrogates test

jdorweiler:
Screen Shot 2021-03-04 at 1 35 25 PM


privacy-protections/surrogates/index.html
```diff
@@ -0,0 +1,37 @@
...
+#results-table th {
+  padding-top: 12px;
+  padding-bottom: 12px;
+  text-align: left;
+  background-color: #f2f2f2;
+  color: black;
+}
+</style>
+</head>
+<body onload="loadSurrogates()">
```
kdzwinel, line 29: can we move this to script to make the linter happy?

privacy-protections/surrogates/main.js
```diff
@@ -0,0 +1,77 @@
...
+
+        document.body.appendChild(s)
+    }
+}   
+
+function updateTable ({name, testData, error}) {
+
+    const table = document.getElementById('results-table')
+    const row = table.insertRow(-1)
+    const testName = row.insertCell(0)
```
kdzwinel, line 19: TIL about `.insertRow` and `.insertCell`. I'm learning!

privacy-protections/surrogates/main.js
```diff
@@ -0,0 +1,77 @@
...
+        test: () => { return window.ga.answer === 42 },
+        cleanUp: () => { delete window.ga }
+    },
+    'google-analytics, ga.js': {
+        url: "https://google-analytics.com/ga.js",
+        test: () => { return !!window._gaq },
+        cleanUp: () => { delete window._gaq }
+    },
+    'Directly accessing a web resouce' : {
+        url: "chrome-extension://lfpgfgegioonagopmelghcelfgffmjnh/web_accessible_resources/analytics.js",
```
kdzwinel, line 71: wait, what extension ID is this? Doesn't look like ours: https://chrome.google.com/webstore/detail/duckduckgo-privacy-essent/bkdgflcldnnnapblkhphbgpggdiikppg ?

privacy-protections/surrogates/main.js
```diff
@@ -0,0 +1,77 @@
...
+        s.src = testData.url
+        
+        s.onload = () => updateTable({name, testData})
+        s.onerror = (error) => updateTable({name, testData, error})
+
+        document.body.appendChild(s)
+    }
+}   
+
+function updateTable ({name, testData, error}) {
```
kdzwinel, line 15: can we also have a global object with test results like in other tests? This way, when testing this page automatically, you will not need to scrape HTML, you'll just pull info from `window.results`.

kdzwinel (changes requested):
Great job with the styling and use of all those ancient table APIs ❤️ That being said, please fix the linting, because

┏┓ 
┃┃╱╲ In this 
┃╱╱╲╲ house 
╱╱╭╮╲╲ we love 
▔▏┗┛▕▔ & appreciate 
╱▔▔▔▔▔▔▔▔▔▔╲ 
Semicolons
╱╱┏┳┓╭╮┏┳┓ ╲╲ 
▔▏┗┻┛┃┃┗┻┛▕▔

I also left some other comments for your enjoyment.

privacy-protections/surrogates/main.js
```diff
@@ -0,0 +1,84 @@
...
+
+(function loadSurrogates () {
+    for (const [name, testData] of Object.entries(surrogates)) {
+        const s = document.createElement('script');
+
+        if (testData.crossOrigin) {
+            s.crossOrigin = testData.crossOrigin;
+        }
+
+        s.src = testData.url;
```
kdzwinel, line 77: I remember that in some cases (local files? cache?) event could fire before listener is set up if you provide url first. Just to be extra sure - can you move this after listeners are ready?

privacy-protections/surrogates/main.js
```diff
@@ -0,0 +1,84 @@
...
+        test: () => { return window.ga.answer === 42; },
+        cleanUp: () => { delete window.ga; }
+    },
+    'google-analytics, ga.js': {
+        url: 'https://google-analytics.com/ga.js',
+        test: () => { return !!window._gaq; },
+        cleanUp: () => { delete window._gaq; }
+    },
+    'Directly accessing a web resouce': {
+        url: 'chrome-extension://bkdgflcldnnnapblkhphbgpggdiikppg/web_accessible_resources/analytics.js',
```
kdzwinel, line 62: I'm still not sure what this extension ID is? Is it random?

kdzwinel: Ah, you changed it to ours - sorry missed that 🤦‍♂️

privacy-protections/surrogates/main.js
```diff
@@ -0,0 +1,84 @@
...
+        testData.cleanup();
+    }
+}
+
+const surrogates = {
+    'google-analytics.com/analytics.js, crossOrigin': {
+        url: 'https://google-analytics.com/analytics.js',
+        crossOrigin: 'anonymous',
+        notes: 'Test loading with crossOrigin set on element (should fail on Firefox) https://bugzilla.mozilla.org/show_bug.cgi?id=1694679',
+        test: () => { return window.ga.answer === 42; },
```
kdzwinel, line 48: why are we testing it this way and not the same way as `_gaq`?

privacy-protections/surrogates/main.js
```diff
@@ -0,0 +1,84 @@
...
+        cleanUp: () => { delete window.ga; }
+    },
+    'google-analytics.com/analytics.js': {
+        url: 'https://google-analytics.com/analytics.js',
+        test: () => { return window.ga.answer === 42; },
+        cleanUp: () => { delete window.ga; }
```
kdzwinel, line 54: we are not waiting for test to load/fail, so all tests are running at once - can there be a race condition when they overlap (e.g. two tests use `window.ga`)?

privacy-protections/surrogates/main.js
```diff
@@ -0,0 +1,84 @@
...
+    },
+    'google-analytics, ga.js': {
```
kdzwinel, line 56: all other tests involve testing different configurations of `analytics.js` surrogate why are we also testing `ga.js` surrogate? Is this some kind of a special case?

kdzwinel (commented):
Code LGTM, I played with it though and realized that I'm missing some of the context so I left some questions.

privacy-protections/surrogates/main.js
jdorweiler: Thanks, fixed it 👀

privacy-protections/surrogates/main.js
jdorweiler: Good point, I learned this would return true with the real ga code too. I updated it to look for an empty object from ga.create()

kdzwinel:
Thanks a lot for answering all my questions!

PR #34, opened 2021-03-28
Add in expensive audio fingerprint lookup

jonathanKingston:


kdzwinel:
WDYT about adding that to https://github.com/duckduckgo/privacy-test-pages/pull/33 instead? We don't rally want to slow this page down and this test doesn't produce any new FP data as far as I understand.

PR #37, opened 2021-05-13
Iframe downloading

jonathanKingston:
Adding in hidden iframes to download something.


kdzwinel (commented):
Nice! Drop the `.download.html.swp` file and feel free to merge 👍

PR #41, opened 2021-05-25
Update observer to check duration and filter failed loads

Charlie-belmer:
This adds additional checks to determine if a FB call actually loads. Unfortunately, there isn't a clean way to determine whether an iFrame loads content successfully. This uses timing as a proxy for that. Should fix integration tests for click to load.


kdzwinel:
Thanks Charlie, it's unfortunate that the original solution doesn't work anymore, but this workaround looks good to me!

PR #42, opened 2021-06-04
Ctl iFrame

Charlie-belmer:
This backs out the previous changes around timing (it was inconsistent), and instead reports the number of iFrames on the page which the browser is expected to request.


kdzwinel:
Thanks Charlie! LGTM.

PR #43, opened 2021-07-15
Check if we can load a surrogate even if the integrity mismatches.

jonathanKingston:
@kdzwinel / @jdorweiler  this currently fails should I set `shouldFail` to true?

The idea here is we would like the browser to permit us loading a surrogate when the SRI check doesn't match. This would allow us to have surrogates for the likes of fingerprint.js and f***Adblock etc.

My hope is we can direct Chrome/Firefox etc at this page when we raise issues.

### DIFF
--- a/privacy-protections/surrogates/main.js
+++ b/privacy-protections/surrogates/main.js
@@ -55,6 +55,15 @@ const surrogates = {
         test: () => { return !!(window.ga && Object.keys(window.ga.create()).length === 0); },
         cleanUp: () => { delete window.ga; }
     },
+    'google-analytics.com/analytics.js broken integrity': {
+        url: 'https://google-analytics.com/analytics.js',
+        crossOrigin: 'anonymous',
+        integrity: 'sha512-1xNTXD/ZeaKg/Xjb6De9la7CXo5gC1lMk+beyKo691KJrjlj0HbZG6frzK0Wo6bm96i9Cp6w/WB4vSN/8zDBLQ==',
+        notes: 'Fails loading in all browsers despite we think it should be possible for the extension to load this.',
+        shouldFail: false,
+        test: () => { return !!(window.ga && Object.keys(window.ga.create()).length === 0); },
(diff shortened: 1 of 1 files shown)


privacy-protections/surrogates/main.js
```diff
@@ -55,6 +55,15 @@ const surrogates = {
         test: () => { return !!(window.ga && Object.keys(window.ga.create()).length === 0); },
         cleanUp: () => { delete window.ga; }
     },
+    'google-analytics.com/analytics.js broken integrity': {
+        url: 'https://google-analytics.com/analytics.js',
+        crossOrigin: 'anonymous',
+        integrity: 'sha512-1xNTXD/ZeaKg/Xjb6De9la7CXo5gC1lMk+beyKo691KJrjlj0HbZG6frzK0Wo6bm96i9Cp6w/WB4vSN/8zDBLQ==',
+        notes: 'Fails loading in all browsers despite we think it should be possible for the extension to load this.',
```
kdzwinel, line 62: How about:

```suggestion
        notes: 'Surrogate will fail to load due to integrity check. We think this check should not apply to extensions.',
```

privacy-protections/surrogates/main.js
```diff
@@ -55,6 +55,15 @@ const surrogates = {
...
+        shouldFail: false,
+        test: () => { return !!(window.ga && Object.keys(window.ga.create()).length === 0); },
```
kdzwinel, line 64: This will be green if surrogate successfully loads, right? It will still be red if extension is disabled and original loads?

jonathanKingston: Yes, although the integrity isn't valid also so shouldn't pass either.

jonathanKingston: AKA this should only pass if the extension redirecting to a local extension resource is carved out within the browser code

kdzwinel:
LGTM! Thanks 👏

PR #49, opened 2021-08-10
Check message origin when retreiving iframe storage results

sammacbeth:
Currently both safe and tracking frames always return the same results because both checks will just consume the first postMessage. This PR adds the missing check to ensure they wait for the message from the correct frame.


kdzwinel:
🤦‍♂️ Thanks for catching it!

PR #50, opened 2021-08-11
Wait for git to finish before restarting the app.

kdzwinel:
We were not waiting for git commands to finish creating a race condition between git and application restart.


server.js
```diff
@@ -53,16 +53,18 @@ app.post('/git', (req, res) => {
...
 
-        console.log('> [GIT] Updated with origin/gh-pages');
+            console.log('> 💚 Deploy successful!');
 
-        return res.sendStatus(200);
+            return res.sendStatus(200);
+        } catch (e) {
+            console.log('> 💥 Error deploying: ', e);
+            return res.sendStatus(500);
+        }
```
sammacbeth, line 67: Why is the catch needed now? Are occasional errors expected when running this sync?

sammacbeth (approved):
Looks good - just wondered why a catch is needed.

PR #53, opened 2021-08-12
Extend surrogate test page to demo iOS issues

sammacbeth:
- Adds info on surrogate load timing - i.e. if the surrogate is already ready on document load, or later.
- Adding `#tamperios` to the URL will also run a script to interfere with surrogate injection on iOS.


kdzwinel:
This feels a bit out of place for this specific test page + it's not documented. If we feel that this page will be useful for a long time let's make a new page with explanation and tampering enabled by default. If we feel that this is something that will not be useful in the long run then let's move to jsbin.

sammacbeth:
Closing as defunct.

PR #54, opened 2021-08-26
Onion skin

jonathanKingston:


privacy-protections/fingerprinting/canvas.js
```diff
@@ -37,12 +37,92 @@ function generateDataURLWithCode (codeInput, alpha = 255) {
...
+                sizeCount[pixelSize] += 1;
+            }
+            const zeroChange = sizeCount.shift();
+            ok(zeroChange === undefined, 'Sanity check on ignored pixels');
+            const singleChange = sizeCount.shift();
+            const changeCount = sizeCount.reduce((a, i) => a + i, 0);
+            const elegiblePixels = singleChange + changeCount;
+            ok(changeCount > 0, 'Should not only have single peturbed changes');
+            // This is set purposefully low to avoid false positives and also our elegible pixel algo is more complex and may change over time.
+            ok(changeCount > (elegiblePixels * 0.2), 'Should have more than 20% of elegiblePixels that are changed across multiple canvases');
```
englehardt, line 113:
```suggestion
            const eligiblePixels = singleChange + changeCount;
            ok(changeCount > 0, 'Should not only have single perturbed changes');
            // This is set purposefully low to avoid false positives and also our eligible pixel algo is more complex and may change over time.
            ok(changeCount > (eligiblePixels * 0.2), 'Should have more than 20% of eligiblePixels that are changed across multiple canvases');
```

privacy-protections/fingerprinting/canvas.js
```diff
@@ -37,12 +37,92 @@ function generateDataURLWithCode (codeInput, alpha = 255) {
...
+                if (!(pixelSize in sizeCount)) {
+                    sizeCount[pixelSize] = 0;
+                }
+                sizeCount[pixelSize] += 1;
+            }
+            const zeroChange = sizeCount.shift();
+            ok(zeroChange === undefined, 'Sanity check on ignored pixels');
+            const singleChange = sizeCount.shift();
+            const changeCount = sizeCount.reduce((a, i) => a + i, 0);
+            const elegiblePixels = singleChange + changeCount;
```
englehardt, line 110: I don't think this is calculated correctly if I understand correctly what you're going for here. Eligible pixels should be a count of the total number of pixels that aren't transparent, whereas this is counting the number of unique values of those pixels. Instead, you can count eligiblePixels in the loop above that checks `shouldIgnorePixel`.

Then if you want to check that at least N% have been changed you also can't use uniques because it might be the case that some pixels were changed to the same value. To count those you can just loop through the final canvas and then increment if the pixel is different than `randData` at the same index.

privacy-protections/fingerprinting/canvas.js
```diff
@@ -37,12 +37,92 @@ function generateDataURLWithCode (codeInput, alpha = 255) {
...
+            ok(changeCount > 0, 'Should not only have single peturbed changes');
+            // This is set purposefully low to avoid false positives and also our elegible pixel algo is more complex and may change over time.
+            ok(changeCount > (elegiblePixels * 0.2), 'Should have more than 20% of elegiblePixels that are changed across multiple canvases');
+            return sizeCount;
```
englehardt, line 114: Why return sizeCount here?

englehardt (changes requested):

privacy-protections/fingerprinting/canvas.js
jonathanKingston: I sum together number of pixels with a single change and the number of pixels that have more than one (for clarity the sizeCount is an array that sums to the complete pixel array size minus the 0th pixel that we ignore). The number is the same but I changed to calculating it in the loop above to simplify the understanding.

I've updated the var names and comments here.

privacy-protections/fingerprinting/canvas.js
jonathanKingston: We expose all the return values to a page global so integration tests could also look at these values and make it's own conclusions.

PR #55, opened 2021-09-09
Simplify canvas layering checks

jonathanKingston:
@englehardt it won't let me add you as a reviewer?! Did you get fired :P

This doesn't look at the RGB distribution but it simplifies how we calculate the differences comparing to the known array and checks the numbers add up to the expected count also.

The maths is the same though the numbers come out the same. I increased the canvas size but kept it to a single check for now as it's pretty costly for perf.


englehardt (approved):
This LGTM. The counts don't change since we're still using uniques (which I think is fine). If we end up adding tests for the distribution across RGB we can also add a check for non-unique changes.

PR #58, opened 2021-09-24
Couple of changes to the surrogates page.

kdzwinel:
- I added test that just checks if surrogate has loaded in the main frame.
- And another one that checks the same thing in an iframe.
- I simplified the table a bit.
- And cleaned up output format to match the standard.
- Removed mention of FF bug because it was fixed - https://bugzilla.mozilla.org/show_bug.cgi?id=1694679 (I left the test, just in case)

surrogates

### DIFF
--- a/privacy-protections/surrogates/frame.html
+++ b/privacy-protections/surrogates/frame.html
@@ -1,0 +1,24 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width">
+  <title>Tracker surrogates</title>
+
+  <script src='https://google-analytics.com/analytics.js' defer></script>
+  <script>
+        const onMessage = msg => {
+            if (msg.data.action && msg.data.action === 'surrogate') {
+                window.parent.postMessage({
+                    surrogate: (window.ga && window.ga.name !== 'N' && window.ga.answer === 42)
+                }, '*');
+            }
+        };
+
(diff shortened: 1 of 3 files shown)


privacy-protections/surrogates/frame.html
```diff
@@ -0,0 +1,24 @@
...
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width">
+  <title>Tracker surrogates</title>
+
+  <script src='https://google-analytics.com/analytics.js' defer></script>
+  <script>
+        const onMessage = msg => {
+            if (msg.data.action && msg.data.action === 'surrogate') {
+                window.parent.postMessage({
+                    surrogate: (window.ga && window.ga.name !== 'N' && window.ga.answer === 42)
```
kdzwinel, line 13: checks if surrogate has loaded and reports back to the main frame.

privacy-protections/surrogates/main.js
```diff
@@ -1,108 +1,148 @@
...
-
     if (testData.cleanUp) {
         testData.cleanUp();
     }
+
+    results.results.push(result);
+}
+
+function checkSurrogate () {
+    if (window.ga && window.ga.name !== 'N' && window.ga.answer === 42) {
```
kdzwinel, line 44: this is a simple way of detecting our surrogate

jdorweiler:
👍

jdorweiler (approved):

PR #59, opened 2021-09-27
Loop protection in HTTPS upgrades

kdzwinel:
Domain is HTTPS upgradable but page will constantly try to downgrade to HTTP. Client should detect that and allow page to load. Unfortunately our extension seems to have a bug and doesn't do it.

https://app.asana.com/0/1186013049913869/1201057222436407/f

### DIFF
--- a/privacy-protections/https-loop-protection/http-only.html
+++ b/privacy-protections/https-loop-protection/http-only.html
@@ -1,5 +1,6 @@
 <!DOCTYPE html>
 <html lang="en">
+
 <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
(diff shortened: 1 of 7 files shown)


privacy-protections/https-loop-protection/index.html
```diff
@@ -0,0 +1,27 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>HTTPS upgrade loop protection</title>
+
+    <script src='./main.js' defer></script>
+    <link href='./style.css' rel='stylesheet'></link>
+</head>
+<body>
+    <p><a href="../../">[Home]</a> ↣ <a href="../">[Privacy Protections Tests]</a> ↣ <strong>[HTTPS Upgrade Loop Protection]</strong></p>
+
+    <p>This tst will navigate to page on a https upgradable domain that imadiatelly redirects to its http version. This will cause a http↔http loop (client trying to upgrade, page downgrading). Clients should detect this scenario and allow page to load.</p>
```
jonathanKingston, line 14:
```suggestion
    <p>This test will navigate to page on a https upgradable domain that immediately redirects to its http version. This will cause a http↔http loop (client trying to upgrade, page downgrading). Clients should detect this scenario and allow page to load.</p>
```

privacy-protections/https-upgrades/main.js
```diff
@@ -174,7 +174,7 @@ function downloadTheResults () {
     const a = document.createElement('a');
     const url = window.URL.createObjectURL(new Blob([data], { type: 'application/json' }));
     a.href = url;
-    a.download = 'fingerprinting-results.json';
+    a.download = 'https-upgrades-results.json';
```
kdzwinel, line 177: unrelated change - renaming downloaded file to match test name

privacy-protections/https-loop-protection/http-only.html
```diff
@@ -0,0 +1,36 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>HTTPS upgrade loop protection</title>
+</head>
+<body>
+    <script>
+        const MAX = 15;
```
kdzwinel, line 10: if loop protection doesn't kick in after 15 attempts - give up

privacy-protections/https-loop-protection/main.js
```diff
@@ -0,0 +1,142 @@
...
+const testsDiv = document.querySelector('#tests');
+const testsSummaryDiv = document.querySelector('#tests-summary');
+const testsDetailsDiv = document.querySelector('#tests-details');
+
+const TEST_DOMAIN = 'good.third-party.site';
+
+const tests = [
+    {
+        id: 'upgrade-navigation',
+        run: () => {
```
kdzwinel, line 13: most of this file, besides the contents of this test, is just a template. We should probably extract it to some reusable helpers instead of copying it over and over

jonathanKingston (approved):
This is really great, thanks for changing the wording to make the test clearer :). Also thanks for the PR comments yourself, I think this helps often.

PR #61, opened 2021-10-01
Test Automation - clean up

kdzwinel:
While writing instructions for test automation I realized there are couple of rough edges that I decided to fix:

- click to play didn't have "results" global variable
- nit: moved click-to-play's JS code from HTML to separate JS file
- some tests were 2 levels deep (e.g. when we had multiple results from an iframe) `results.results[0].value[0].value` - I flattened them (so that all tests are `results.results[0].value`)
- some tests were returning error message as a value - I changed that so that value is null when test fails and error is in separate property
- some tests had value in `result` property while the standard is `value`


privacy-protections/click-to-load/index.html
```diff
@@ -56,88 +56,7 @@ <h2> Objects in iFrames </h2>
...
-    a.click();
-
-    window.URL.revokeObjectURL(url);
-    a.remove();
-}
-
-const downloadButton = document.querySelector('#download');
-downloadButton.addEventListener('click', () => downloadTheResults());
-</script>
+<script src='./main.js'></script>
```
kdzwinel, line 59: moved JS code to separate file

privacy-protections/click-to-load/main.js
```diff
@@ -0,0 +1,100 @@
+let facebookCalls = 0;
+let fbIFrames = 0;
+
+// object that contains results of all tests
+const results = {
```
kdzwinel, line 5: this is new, click-to-play didn't have global variable before

jonathanKingston (approved):

PR #63, opened 2021-10-20
Add delayed tracker test cases

jonathanKingston:
I added two cases (where we're using script to inject trackers) to delay loading of the tracker if the ?delay=1000 etc param is passed.

This allows us to verify if the reporting in mobile updates when the tracker is blocked etc. I notice we don't do the animation but clicking on the chip displays the number as and when the blocking happens.

Test case for: https://app.asana.com/0/414235014887631/1201210452027324/f


package-lock.json
```diff
@@ -1,8 +1,2676 @@
 {
   "name": "privacy-test-pages",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
kdzwinel, line 4: oh no, not this again. If we are switching to lockfile v2 we should add a hard requirement for new npm/node (14+?)

package-lock.json
```diff
@@ -1,8 +1,2676 @@
...
   "requires": true,
+  "packages": {
+    "": {
+      "name": "privacy-test-pages",
```
kdzwinel, line 8: so weird

tracker-reporting/document-fragment.html
```diff
@@ -5,46 +5,30 @@
...
 
+function loadTracker () {
     const targetElement = document.getElementById('target-div');
 
     const markup = "<img src='//facebook.com/tr?test=1'>";
 
     const fragment = document.createDocumentFragment();
     const div = document.createElement('div');
     div.innerHTML = markup;
-    // var d = _.map(c.getElementsByTagName('script'), _.identity);
```
kdzwinel, line 36: nice 🧼

tracker-reporting/utils.js
```diff
@@ -0,0 +1,10 @@
+// Delays a function in ms based on a URL param delay=x
+// eslint-disable-next-line no-unused-vars
+function callAfterDelay (callback) {
+    if (location.search.includes('delay')) {
+        const delay = location.search.match(/[&?]delay=([0-9]+)/);
+        setTimeout(callback, delay[1]);
```
kdzwinel, line 6: Let's make this code POP by using SearchParams.

```suggestion
    const params = (new URL(location.href)).searchParams;
    const delay = parseInt(params.get('delay'), 10) || 0;

    if (delay) {
        setTimeout(callback, delay);
```

PR #65, opened 2021-11-08
Add autoconsent test page

sammacbeth:
This is a basic test page for consent popup management.

### DIFF
--- a/features/autoconsent/index.html
+++ b/features/autoconsent/index.html
@@ -1,0 +1,38 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width">
+  <title>Autoconsent Test</title>
+
+  <style>
+      /* styles, if needed */
+  </style>
+  <script src='./main.js' defer></script>
+</head>
+<body>
+<p><a href="/index.html">[Home]</a></p>
+  
+<p>Tests for automatic consent popup clicking.</P>
+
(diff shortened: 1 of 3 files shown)


features/autoconsent/index.html
```diff
@@ -0,0 +1,38 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width">
+  <title>Autoconsent Test</title>
+
+  <style>
+      /* styles, if needed */
+  </style>
```
kdzwinel, line 10:
```suggestion
```

🧼

features/autoconsent/main.js
```diff
@@ -0,0 +1,82 @@
+// populate test sites
+const testSites = {
```
kdzwinel, line 2: why this format is easier than keeping it in HTML? We use this format elsewhere and it's easier to update in the future?

kdzwinel (approved):
Thanks @sammacbeth ! Looks great, but please link this test page from index.html + I left one more comment. Feel free to merge after that.

features/autoconsent/main.js
sammacbeth: This way is less verbose, and allows test-cases to be easily copied over to/from the tests in the autoconsent project: https://github.com/cliqz-oss/autoconsent/blob/master/addon/test.ts#L82-L91

PR #66, opened 2021-11-26
Move autoconsent button creation to script

sammacbeth:
This fixes a race where the button is clicked before the event listener is attached.


kdzwinel (approved):
LGTM

PR #67, opened 2021-12-06
Query Parameters Test Page

SlayterDev:
This PR adds a test page for query parameters. Clicking links will take you to a landing page and display the supplied parameters. On platforms that support it you can use this to verify parameter stripping is working.

### DIFF
--- a/features/autoconsent/index.html
+++ b/features/autoconsent/index.html
@@ -1,0 +1,34 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width">
+  <title>Autoconsent Test</title>
+
+  <script src='./main.js' defer></script>
+</head>
+<body>
+<p><a href="/index.html">[Home]</a></p>
+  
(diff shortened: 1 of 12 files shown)


kdzwinel:
Looks good! But we should ideally automate those pages a bit more.

Similarly to https://privacy-test-pages.glitch.me/privacy-protections/referrer-trimming/ we could have one "start" button that opens couple of pages and reports back how url has changed. In fact I think it'd be simpler than referrer trimming because if we open new tab from same domain with `window.open` we can get it's url quite easily back:

```js
const win = window.open('http://localhost:3000/privacy-protections/query-parameters/query.html?utm_source=something&q=other')
// wait for it to load
// get url
win.location.href
```

If this won't work for some reason on iOS then you require clicks on links but they can automatically report back to the original page the result (window.opener) and close themselves to make it easier to test.

Once we haver results we should expose them via global object.

LMK if anything is unclear!

kdzwinel (approved):
Lets merge as discussed and add automation as a follow up 👍

PR #68, opened 2021-12-10
Add YouTube test page

kzar:
There are lots of different types of embedded YouTube videos and ways
of controlling them using the YouTube Iframe API[1]. Adding a test
page so that we can easily test they all work correctly with our
extension and browsers.

1 - https://developers.google.com/youtube/iframe_api_reference

### DIFF
--- a/index.html
+++ b/index.html
@@ -57,6 +57,7 @@
 	<li><a href='./privacy-protections/https-upgrades/'>HTTPS upgrades</a></li>
 	<li><a href='./privacy-protections/https-loop-protection/'>HTTPS upgrade loop protection</a></li>
 	<li><a href='./privacy-protections/click-to-load/'>Facebook click to load</a></li>
+	<li><a href='./privacy-protections/youtube-click-to-load/'>YouTube click to load</a></li>
 	<li><a href='./privacy-protections/surrogates/'>Surrogates</a></li>
 	<li><a href='./privacy-protections/gpc/'>Global Privacy Control</a></li>
 	<li><a href='./privacy-protections/amp/'>AMP Links</a></li>
--- a/privacy-protections/youtube-click-to-load/index.html
+++ b/privacy-protections/youtube-click-to-load/index.html
@@ -1,0 +1,257 @@
+<!doctype html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width">
+
+  <title>YouTube click to load</title>
+
+  <style>
+   div { border: 2px #000 solid; padding: 10px; margin: 10px; }
+   #short-container { height: 200px; overflow: hidden; }
+   .spherical-control { width: 80px; }
+  </style>
+</head>
+<body>
+  <header>
+    <p>
+      <a href="../../">[Home]</a> ↣
+      <a href="../">[Privacy Protections Tests]</a> ↣
+      <strong>[YouTube click to load]</strong>
+    </p>
+    <h1>YouTube click to load tests</h1>
+    <p>
+      Different types of embedded YouTube videos and playlists. Some are created
+      and/or controlled using <a href="https://developers.google.com/youtube/iframe_api_reference">the YouTube Iframe API</a>.
+    </p>
+    <ul>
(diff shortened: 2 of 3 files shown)


kzar:
I don't have permission to assign you, but want to take a look @kdzwinel ?

index.html
```diff
@@ -56,7 +56,8 @@ <h2>Privacy Protections Tests</h2>
 	<li><a href='./privacy-protections/referrer-trimming/'>Referrer trimming</a></li>
 	<li><a href='./privacy-protections/https-upgrades/'>HTTPS upgrades</a></li>
 	<li><a href='./privacy-protections/https-loop-protection/'>HTTPS upgrade loop protection</a></li>
-	<li><a href='./privacy-protections/click-to-load/'>Facebook click to load</a></li>
+	<li><a href='./privacy-protections/facebook-click-to-load/'>Facebook click to load</a></li>
```
kdzwinel, line 59: sorry, I'm afraid you can't change that URL because some integration tests (like in our extension) already depend on it. So our options are:

- make a redirect in server.js so that /click-to-load/ loads /facebook-click-to-load/
- keep facebook where it is now
- move youtube to  a subfolder in /click-to-load/ ? Not sure about this one, but maybe it makes sense?

privacy-protections/youtube-click-to-load/index.html
```diff
@@ -0,0 +1,243 @@
...
+    </p>
+    <h1>YouTube click to load tests</h1>
+    <p>
+      Different types of embedded YouTube videos and playlists. Some are created
+      and/or controlled using <a href="https://developers.google.com/youtube/iframe_api_reference">the YouTube Iframe API</a>.
+      Behaviour of the videos should remain the same when the extension is
+      enabled (after they are clicked to load), except where noted.
+    </p>
+    <p>
+      When testing, ensure that requests to YouTube are blocked untill the first
```
kdzwinel, line 30:
```suggestion
      When testing, ensure that requests to YouTube are blocked until the first
```

privacy-protections/youtube-click-to-load/index.html
```diff
@@ -0,0 +1,243 @@
...
+      video is clicked to load. Note that requests to <u>apis.google.com</u> are not
+      expected to be blocked.
+    </p>
+    <p>
+  </header>
+  <main>
+    <div id="short-container">
+      <p>wwww.youtube.com embedded video.</p>
```
kdzwinel, line 38:
```suggestion
      <p>www.youtube.com embedded video.</p>
```

kdzwinel (commented):
Nice work! I left some comments + can we add self-evaluation like Charlie did for facebook? So that page tries to figure out, as much as possible, and report back on status of those iframes? We may be able to figure out what has loaded using [resource timing api](https://developer.mozilla.org/en-US/docs/Web/API/Resource_Timing_API/Using_the_Resource_Timing_API) and e.g. check if calls were upgraded to youtube-nocookie.com.

index.html
kzar: Oops, I didn't realise that sorry. I've moved it back for now, we can always rethink if it becomes problematic.

privacy-protections/youtube-click-to-load/index.html
kzar: Done.

privacy-protections/youtube-click-to-load/index.html
kzar: Done.

kzar:
Good point about automating the tests. I'm not exactly sure how that would work with videos, but I will give it some thought. Would you mind if I did that as a follow-up though?

kdzwinel (approved):
Yeah, that's OK, but please remember to add it before you wrap up your project 👍

kzar:
@kdzwinel OK cool, I have adjusted the page to trigger [that CORS bug on Firefox < 89](https://bugzilla.mozilla.org/show_bug.cgi?id=1694679) and also improved the copy. Any chance you could take another pass?

kdzwinel:
LGTM, thanks!

PR #69, opened 2021-12-15
Remove trigger for CORS bug in YouTube testpage

kzar:
It turns out that `crossorigin="anonymous"` prevents
https://www.youtube.com/iframe_api from loading, even without our
extension installed. There's no need for us to handle the CORS bug[1] in
older versions of Firefox therefore.

1 - https://bugzilla.mozilla.org/show_bug.cgi?id=1694679


kdzwinel (approved):
No worries, I missed it also :)

PR #70, opened 2021-12-16
Adding tests for the presence of `navigator.ddg`

shakyShane:
This is a WIP, opening early for feedback

https://app.asana.com/0/72649045549333/1201487308673313/f

### DIFF
--- a/features/navigator-namespace.html
+++ b/features/navigator-namespace.html
@@ -1,0 +1,40 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width">
+    <title>Navigator Namespace</title>
+</head>
+<body>
+<p><a href="../index.html">[Home]</a></p>
+
+<p>Using the new DOM API, <code>navigator.ddg</code>, to provide signals about the presence of DDG products</p>
+
+<p id="ddg"></p>
+<p id="ddg-extension"></p>
+
+<script>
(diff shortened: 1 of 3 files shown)


kdzwinel (commented):
Lookin' good! One nice to have idea - can you export results of the test to the `results` object like we do for privacy features? We may want to write an integration test against this page and having that object makes it easier to process the results. This can be as simple as

```js
window.results = {
    page: 'navigator-namespace',
    date: (new Date()).toUTCString(),
    results: [{id: 'namespace', value: true},…]
}
```

shakyShane:
@kdzwinel updated to suit your comment.

features/navigator-namespace.html
```diff
@@ -0,0 +1,40 @@
...
+
+    if (hasNamespace) {
+        if (typeof navigator.ddg.isExtension === 'function') {
+            isExtension = navigator.ddg.isExtension();
+        }
+    }
+
+    window.results = {
+        page: 'navigator-namespace',
+        date: null,
```
kdzwinel, line 28:
```suggestion
        date: (new Date()).toUTCString(),
```

let's have a date in case someone stores the results

kdzwinel (approved):
Left one nit, feel free to merge after this. Thanks!

shakyShane:
@kdzwinel Just after this merged, I realised there's a slight race condition with the way the extension loads content-scripts - meaning this test will never actually pass. I'm sending a follow up to poll for a second before it concludes with a result

PR #71, opened 2021-12-17
Introduce polling to give the extension chance to inject content-scripts

shakyShane:
Sorry for the fast-follow PR, but this should make future tests far more deterministic

https://app.asana.com/0/72649045549333/1201487308673313/f

### DIFF
--- a/features/navigator-interface.html
+++ b/features/navigator-interface.html
@@ -3,7 +3,7 @@
 <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width">
-    <title>Navigator Namespace</title>
+    <title>Navigator Interface</title>
 </head>
 <body>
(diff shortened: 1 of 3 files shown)


shakyShane:
I also updated the naming to match the privacy configuration changes, which are now live

kdzwinel:
no worries, thanks for a follow up. I don't want to bake extension's limitations into tests by default, ideally this API should be always available. Let's reuse https://github.com/duckduckgo/privacy-test-pages/blob/33ee5f0772ba9b58e753063f78422b26e6133397/tracker-reporting/utils.js which allows us to pass delay value as a test param.

shakyShane:
@kdzwinel I matched the style of the other tests that use `callAfterDelay` - hope this suits

PR #72, opened 2021-12-21
change from `ddg` to `duckduckgo` on the navigator interface

shakyShane:
Following some feedback from Product [here](https://app.asana.com/0/0/1201508595219526/1201542311746546/f) it was decided to switch to `navigator.duckduckgo` instead of `navigator.ddg` - this PR reflects that.

### DIFF
--- a/features/navigator-interface.html
+++ b/features/navigator-interface.html
@@ -12,13 +12,17 @@
             callAfterDelay(detectInterface);
         }
 
-        function detectInterface () {
-            const hasInterface = 'ddg' in navigator;
-            let isExtension = false;
+        async function detectInterface () {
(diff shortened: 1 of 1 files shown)


jonathanKingston (changes requested):
We had some discussion about changing the isExtension interface to be:

```
navigator.duckduckgo.platform
```

We should probably return that `.platform` value in the results.

I think we should block this PR on that, but otherwise is great :)

shakyShane:
@jonathanKingston @kdzwinel Please see the new changes in this PR, it should address the previous concerns

jonathanKingston (approved):
This looks pretty robust to me, thanks for making those changes!

PR #74, opened 2022-01-06
[WIP] Implement test pages for platform-level privacy impacting APIs

englehardt:
See: https://app.asana.com/0/72649045549333/1201495751273690/f


privacy-protections/storage-partitioning/helpers/commonTests.js
```diff
@@ -0,0 +1,552 @@
...
+//    cross-site: [ { "value": "fd10f7f5d2cf" }, { "value": "fd10f7f5d2cf" } ]
+function validateStorageAPI (sameSites, crossSites, reference, random) {
+    if (
+        (sameSites.every(v => v.error === 'Unsupported')) &&
+        (crossSites.every(v => v.error === 'Unsupported'))
+    ) {
+        return 'unsupported';
+    } else if (
+        (sameSites.every(v => v.error && v.error.endsWith('is deprecated'))) &&
+        (crossSites.every(v => v.error && v.error.endsWith('is deprecated')))
```
kdzwinel, line 47: not sure if I'm reading it correctly, but if API is deprecated (e.g. websql) it still can be used and we don't know when this will go away, so I would ignore the fact that it's deprecated in testing

privacy-protections/storage-partitioning/helpers/commonTests.js
```diff
@@ -0,0 +1,552 @@
...
+        }
+        return 'fail';
+    }
+
+    if (
+        // (!sameSites.length === configurations['same-site'].iterations) ||
+        // (!crossSites.length === configurations['cross-site'].iterations) ||
+        (!sameSites.every(v => v.value === reference.value)) ||
+        (!crossSites.every(v => v.value === crossSites[0].value)) ||
+        (!crossSites.every(v => v.value !== reference.value))
```
kdzwinel, line 64: I understand that having same value cross site is a failure (no partitioning), but why having same value across `sameSites` is a failure? (I might be reading those files in the wrong order)

kdzwinel (commented):
Nicely done!I did a first quick read-through. I'll post more comments later today.

privacy-protections/storage-partitioning/helpers/commonTests.js
```diff
@@ -0,0 +1,552 @@
...
+        store: (data) => {
+            // we are using same set of tests for main frame and an iframe
+            // we want to set 'Lax' in the main frame for the cookie not to be suspicious
+            // we want to set 'None' in an iframe for cookie to be accessible to us
+            const sameSite = (window !== window.top) ? 'None' : 'Lax';
+
+            document.cookie = `jsdata=${data}; expires= Wed, 21 Aug 2030 20:00:00 UTC; Secure; SameSite=${sameSite}`;
+        },
+        retrieve: () => {
+            return document.cookie.match(/jsdata=([0-9a-z-]+)/)[1];
```
kdzwinel, line 111: we alrady have a cookie named `jsdata` for https://privacy-test-pages.glitch.me/privacy-protections/storage-blocking/ test, I'd suggest other name for those cookies not to interfere. Perhaps we should use a name that makes it clear form which test cookie is comming

privacy-protections/storage-partitioning/helpers/commonTests.js
```diff
@@ -0,0 +1,552 @@
...
+        },
+        validate: validateStorageAPI
+    },
+    localStorage: {
+        type: 'storage',
+        store: (data) => {
+            localStorage.setItem('data', data);
```
kdzwinel, line 118: same here, we are already using 'data'. Maybe let's use something that includes test page name across the board?

privacy-protections/storage-partitioning/helpers/commonTests.js
```diff
@@ -0,0 +1,552 @@
...
+    },
+    'Cache API': {
+        type: 'storage',
+        store: (data) => {
+            return caches.open('data').then((cache) => {
+                const res = new Response(data, {
+                    status: 200
+                });
+
+                return cache.put('/cache-api-response', res);
```
kdzwinel, line 190: let's keep it in the `/partitioning` namespace with other stuff maybe to avoid conflicts with other tests?

privacy-protections/storage-partitioning/main.js
```diff
@@ -0,0 +1,253 @@
...
+    testURL.searchParams.set('sessionId', sessionId);
+    testURL.searchParams.set('isLocalTest', isLocalTest);
+
+    // Ideally we'd use noopener here as some browsers (e.g., Firefox) keep tabs that
+    // have references to each other in the same "agent cluster", which impacts the
+    // scope of storage. This is why you'll see consistent sessionStorage across tabs
+    // that have references to each other in Firefox. Unfortuantely the test tab can't
+    // be closed programatically unless it has an opener reference.
+    // See: https://textslashplain.com/2021/02/04/window-close-restrictions/
+    window.open(testURL, '_blank');
```
kdzwinel, line 138: I remember some browsers considering to assume 'noopener' for `target=_blank`. You haven't observed it though, right?

privacy-protections/storage-partitioning/main.js
```diff
@@ -0,0 +1,253 @@
...
+        toggleDetailsButton.innerText = 'Show Detailed Results';
+        addUnhideHandler();
+    }, { once: true });
+}
+addUnhideHandler();
+
+// run tests if button was clicked or…
+runButton.addEventListener('click', () => runTests());
+
+// if url query is '?store' store the data immadiatelly
```
kdzwinel, line 243:
```suggestion
// if url query is '?run' run the tests immediately
```

privacy-protections/storage-partitioning/main.js
```diff
@@ -0,0 +1,253 @@
...
+if (document.location.search === '?run') {
+    runTests();
+}
+
+// warn the user if loaded over localhost
+if (window.location.hostname === 'localhost') {
```
kdzwinel, line 249: nice touch!

privacy-protections/storage-partitioning/server/routes.js
```diff
@@ -0,0 +1,173 @@
...
+    prefetch: 'text/html',
+    css: 'text/css',
+    xhr: 'text/html'
+};
+
+const countCache = new NodeCache({ stdTTL: cacheTTL });
+const cacheKey = (key, fileType) => `${key}_${fileType}`;
+
+router.get('/resource', (req, res) => {
+    console.log(req.hostname);
```
kdzwinel, line 45: can you please clean up unnecessary logs?

.gitignore
```diff
@@ -1,2 +1,4 @@
 node_modules
-.DS_Store
\ No newline at end of file
+.DS_Store
+.*.sw*
```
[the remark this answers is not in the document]
englehardt: Vim swap files (I also deleted one that was accidentally added to the repo)

LICENSE.md
```diff
@@ -0,0 +1,52 @@
+Copyright (c) 2010-2022 Duck Duck Go, Inc.
```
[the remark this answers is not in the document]
englehardt: Yes. I need to update the license to make it clear that we're including code from privacytests.org. I'll update this file and add headers to the files that include third-party code.

README.md
```diff
@@ -30,7 +30,47 @@ We have couple of test domains, that all resolve to `privacy-test-pages.glitch.m
...
+Unfortunately neither of these approaches support wildcard subdomains, so you will need to add new subdomains as required by your tests.
+
+##### Adding HTTPS support for test domains
+On MacOS:
+```
+brew install mkcert
+brew install nss # if you use Firefox
+```
+
+Next, run the following command to make your OS cert store and Firefox's cert store trust your cert:
```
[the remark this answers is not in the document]
englehardt: Yes. The other browsers use the OS's cert store and mkcrt will add the root there. I think nss is just required to interact with Firefox's cert store. (FYI these steps are from https://web.dev/how-to-use-local-https/)

privacy-protections/storage-partitioning/.eslintrc
```diff
@@ -0,0 +1,12 @@
+{
+    "globals": {
+        "commonTests": "writable",
+        "DB": "writable"
+    },
+    "parserOptions": {
+        "sourceType": "script",
+        "ecmaFeatures": {
+            "globalReturn": false
```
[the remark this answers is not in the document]
englehardt: Admittedly I copied this from the storage blocking folder and haven't updated it / haven't thought to deeply about it. Let me do that and get back to you :)

privacy-protections/storage-partitioning/helpers/idb-wrapper.js
```diff
@@ -0,0 +1,69 @@
+// https://raw.githubusercontent.com/msandrini/minimal-indexed-db/master/index.js
```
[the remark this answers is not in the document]
englehardt: Will do! I just copied things over from storage blocking and hoped to clean it up before flagging you for review.

privacy-protections/storage-partitioning/helpers/uuidv4.min.js
```diff
@@ -0,0 +1,3 @@
+/* eslint-disable */
```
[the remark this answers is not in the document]
englehardt: Only available in Tech Preview in Safari :(

privacy-protections/storage-partitioning/server/routes.js
```diff
@@ -0,0 +1,173 @@
...
+const resourceFiles = {
+    favicon: 'favicon.png',
+    fetch: 'page.html',
+    font: 'font.woff',
+    image: 'image.png',
+    page: 'page.html',
+    preload: 'page.html',
+    prefetch: 'page.html',
+    xhr: 'page.html'
+};
```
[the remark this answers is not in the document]
englehardt: Thanks! I should de-dupe whatever I can here.

privacy-protections/storage-partitioning/helpers/commonTests.js
englehardt: Good catch! I'll take another look at this. I had assumed that a value wasn't returned, but I'll double check. See: https://github.com/duckduckgo/privacy-test-pages/blob/4fd601c3ac15cc6d50812b0a7ef68e16f3d39ef6/privacy-protections/storage-partitioning/helpers/commonTests.js#L153-L158

privacy-protections/storage-partitioning/helpers/commonTests.js
englehardt: Line 62 will be true if any value returned for the same-site calls isn't equal to the reference value (read in a same-site iframe in the same tab where the values were set). i.e., it will be true if reference = 123, and the values read in the same-site iframes are [123, 456]. Thinking through it, I don't think we want to fail in this condition and should instead return error.

I should add this in a comment, but my intention with the results is:
* API unsupported --> "unsupported"
* An exact set of conditions we expect when partitioned/blocked --> "pass" (ideally this is "blocked" or "partitoned" in the future)
* An exact set of conditions we expect when unpartitioned/unblocked --> "fail"
* Some unknown / unexpected combination of conditions --> "error"

Line 62 returning true should never happen (partitioned or not), and thus should be an error.

privacy-protections/storage-partitioning/main.js
englehardt: I'll verify. Arthur and I found that noopener explains some of the differences we see for Chrome re: https://app.asana.com/0/0/1201683514701538/f, so it might make sense to have two versions of the tests--once which preserves the opener and one which severs it.

privacy-protections/storage-partitioning/helpers/commonTests.js
englehardt: Looks like this is no longer necessary to special case.

privacy-protections/storage-partitioning/helpers/commonTests.js
englehardt: I tightened up the logic quite a bit here and documented it.

privacy-protections/storage-partitioning/main.js
englehardt: I haven't observed this, and decided not to add the noopener version of the test for now. I'll make a not in Asana to consider it.

privacy-protections/storage-partitioning/.eslintrc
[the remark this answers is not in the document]
englehardt: Done.

privacy-protections/storage-partitioning/helpers/idb-wrapper.js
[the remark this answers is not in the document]
englehardt: Moved this to the top-level helpers directory so it can be shared by both test pages (and more).

privacy-protections/storage-partitioning/server/routes.js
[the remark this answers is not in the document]
englehardt: I've decided against de duplicating for now. I think the server itself needs refactoring to move the routing logic into individual test directories, like I've done here. When we do that we can create a central dummy asset folder as well.

privacy-protections/storage-partitioning/iframe.html
```diff
@@ -0,0 +1,16 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Storage blocking test page - iframe</title>
+    <script src='/helpers/idb-wrapper.js'></script>
+    <script src='./helpers/common.js'></script>
+    <script src='./helpers/tests.js'></script>
+    <script src='./iframe.js'></script>
+</head>
+<body>
+    <p>Iframe Test Page</p>
+    <span id="css"></span>
```
kdzwinel, line 14: nit: I don't think that's used?

privacy-protections/storage-partitioning/index.html
```diff
@@ -0,0 +1,39 @@
...
+    <link id="favicon" rel="shortcut icon" type="image/png" href="" />
+</head>
+<body>
+    <p><a href="../../">[Home]</a> ↣ <a href="../">[Privacy Protections Tests]</a> ↣ <strong>[Storage Partitioning Test Page]</strong></p>
+
+    <p>This page will determine whether various storage APIs are parititoned or blocked across origins.</p>
+    <p>NOTE: These tests will return incorrect results if:
+        <ul>
+            <li>you have multiple copies of the test page open</li>
+            <li>you run the tests after force reloading the page, as this <a href="https://web.dev/service-worker-lifecycle/#shift-reload">bypasses</a> Service Workers</li>
```
kdzwinel, line 23: I also get different results with devtools open in Chrome, so we can note that if you can reproduce

kdzwinel, line 23: I also get different results with devtools open in Chrome, so we can note that if you can reproduce

privacy-protections/storage-partitioning/main.js
```diff
@@ -0,0 +1,263 @@
...
+            headers: {
+                Accept: 'application/json, text/plain, */*',
+                'Content-Type': 'application/json'
+            }
+        });
+        if (resp.status !== 200) {
+            console.error(`Error fetching final results ${resp.status}.`);
+            return;
+        }
+        const results = JSON.parse(await resp.text());
```
kdzwinel, line 182: I think you can use `resp.json()` here?

kdzwinel, line 182: I think you can use `resp.json()` here?

privacy-protections/storage-partitioning/serviceworker.js
```diff
@@ -0,0 +1,28 @@
...
+    self.skipWaiting();
+});
+
+self.addEventListener('activate', function (event) {
+    event.waitUntil(self.clients.claim());
+});
+
+self.addEventListener('fetch', async (event) => {
+    const scope = self.registration.scope;
+    const shortPath = event.request.url.split(scope)[1];
```
kdzwinel, line 19: nit: I'd put `new URL(event.request.url)` into a variable and reused it here (url.pathname) and below.

privacy-protections/storage-partitioning/helpers/common.js
```diff
@@ -0,0 +1,42 @@
+/* exported THIRD_PARTY_HOSTNAME THIRD_PARTY_HTTP THIRD_PARTY_HTTPS FIRST_PARTY_HOSTNAME FIRST_PARTY_HTTP FIRST_PARTY_HTTPS accessStorageInIframe */
+const isLocalTest = window.location.hostname.endsWith('.example');
+
+const THIRD_PARTY_HOSTNAME = isLocalTest ? 'third-party.example' : 'good.third-party.site';
+const THIRD_PARTY_HTTP = isLocalTest ? `http://${THIRD_PARTY_HOSTNAME}:3000` : `http://${THIRD_PARTY_HOSTNAME}`;
+const THIRD_PARTY_HTTPS = `https://${THIRD_PARTY_HOSTNAME}:443`;
```
kdzwinel, line 6: I don't think the port is needed?

privacy-protections/storage-partitioning/helpers/common.js
```diff
@@ -0,0 +1,42 @@
...
+
+const FIRST_PARTY_HOSTNAME = isLocalTest ? 'first-party.example' : 'www.first-party.site';
```
kdzwinel, line 8: www.first-party.site doesn't resolve?

kdzwinel (commented):
Phew, this grew BIG!

I tested locally - no issues found, reviewed the code - found some nits and left couple questions. Other than that it looks good to go 🚀

privacy-protections/storage-partitioning/index.html
englehardt: I'm not able to reproduce. Let's see if it persists once the pages are live and we can add a note.

privacy-protections/storage-partitioning/helpers/common.js
englehardt: Fixed. I just needed to add it in Glitch.

PR #75, opened 2022-01-25
Add false positive AMP urls

SlayterDev:
This PR adds some false positive AMP urls. Urls that should trigger the AMP algorithm but are not actual AMP pages.


kdzwinel (approved):
Looks good!

PR #76, opened 2022-02-28
Add cookieStore API

sammacbeth:
The [CookieStore API](https://developer.mozilla.org/en-US/docs/Web/API/CookieStore) is a new API for getting and setting cookies.

This PR adds this API to the set of storage blocking tests.


kdzwinel (approved):
Thanks Sam! Sorry it took forever. BTW crawler now collects CookieStore usage data.

PR #78, opened 2022-03-16
Harmful APIs test page

kdzwinel:
This page will test if web APIs that are considered harmful are available in your browser. Only availablity is verified, privacy concerns might have been mitigated.

### DIFF
--- a/features/harmful-apis/main.js
+++ b/features/harmful-apis/main.js
@@ -1,0 +1,329 @@
+const PAGE_NAME = 'harmful-apis';
+
+const startButton = document.querySelector('#start');
+const downloadButton = document.querySelector('#download');
+
+const testsDiv = document.querySelector('#tests');
+const testsSummaryDiv = document.querySelector('#tests-summary');
+const testsDetailsDiv = document.querySelector('#tests-details');
+
+const tests = [
+    {
+        id: 'battery-api',
+        run: () => {
+            return ('getBattery' in navigator);
+        }
+    },
+    {
+        id: 'webbluetooth-api',
+        run: () => {
+            return ('bluetooth' in navigator);
+        }
+    },
+    {
+        id: 'webusb-api',
+        run: () => {
+            return ('usb' in navigator);
+        }
+    },
+    {
+        id: 'webnfc-api',
+        run: () => {
+            return ('NDEFReader' in window);
+        }
+    },
+    {
+        id: 'webserial-api',
+        run: () => {
+            return ('serial' in navigator);
+        }
+    },
+    {
+        id: 'webhid-api',
+        run: () => {
+            return ('hid' in navigator);
+        }
+    },
+    {
+        id: 'accelerometer-api',
+        run: () => {
+            return ('Accelerometer' in window);
+        }
+    },
+    {
+        id: 'linearaccelerationsensor-api',
+        run: () => {
+            return ('LinearAccelerationSensor' in window);
+        }
+    },
+    {
+        id: 'absoluteorientationsensor-api',
+        run: () => {
+            return ('AbsoluteOrientationSensor' in window);
+        }
+    },
+    {
+        id: 'gyroscope-api',
+        run: () => {
+            return ('Gyroscope' in window);
+        }
+    },
+    {
+        id: 'relativeorientationsensor-api',
+        run: () => {
+            return ('RelativeOrientationSensor' in window);
+        }
+    },
+    {
+        id: 'gravitysensor-api',
+        run: () => {
+            return ('GravitySensor' in window);
+        }
+    },
+    {
+        id: 'ambientlightsensor-api',
+        run: () => {
+            return ('AmbientLightSensorSensor' in window);
+        }
+    },
+    {
+        id: 'magnetometer-api',
+        run: () => {
+            return ('Magnetometer' in window);
+        }
+    },
+    {
+        id: 'keyboard-api',
+        run: () => {
+            return ('keyboard' in navigator);
+        }
+    },
+    {
+        id: 'installed-related-apps-api',
+        run: () => {
+            return ('getInstalledRelatedApps' in navigator);
+        }
+    },
+    {
+        id: 'hardwareConcurrency-api',
+        run: () => {
+            return ('hardwareConcurrency' in navigator);
+        }
+    },
+    {
+        id: 'deviceMemory-api',
+        run: () => {
+            return ('deviceMemory' in navigator);
+        }
+    },
+    {
+        id: 'mediadevices-enumarete-devices-api',
+        run: () => {
+            return ('MediaDevices' in window) && ('enumerateDevices' in MediaDevices.prototype);
+        }
+    },
+    {
+        id: 'storage-manager-estimate-api',
+        run: () => {
+            return ('StorageManager' in window) && ('estimate' in StorageManager.prototype);
+        }
+    },
+    {
+        id: 'network-information-api',
+        run: () => {
+            return ('connection' in navigator);
+        }
+    },
+    {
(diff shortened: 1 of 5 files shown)


features/harmful-apis/main.js
```diff
@@ -0,0 +1,329 @@
...
+    {
+        id: 'gravitysensor-api',
+        run: () => {
+            return ('GravitySensor' in window);
+        }
+    },
+    {
+        id: 'ambientlightsensor-api',
+        run: () => {
+            return ('AmbientLightSensorSensor' in window);
```
englehardt, line 86: This should be `AmbientLightSensor`?

englehardt, line 86:
```suggestion
            return ('AmbientLightSensor' in window);
```

features/harmful-apis/main.js
```diff
@@ -0,0 +1,329 @@
...
+        id: 'push-api',
+        run: () => {
+            return ('ServiceWorkerRegistration' in window) && ('pushManager' in ServiceWorkerRegistration.prototype);
+        }
+    },
+    {
+        id: 'first-party-sets-api',
+        run: () => {
+            if (!window.cookieStore) {
+                return '❌ Can\'t be tested - CookieStore not available';
```
englehardt, line 218: This results in some logic errors when the result are parsed and displayed because it's not `=== false`. It also messed up the downloaded results.

On FF and Safari you'll see:
```
first-party-sets-api - ✅ API available ("❌ Can't be tested - CookieStore not available")
```

features/harmful-apis/main.js
```diff
@@ -0,0 +1,329 @@
...
+
+// object that contains results of all tests
+const results = {
+    page: `${PAGE_NAME}-test`,
+    date: null,
+    results: []
+};
+
+function resultToHTML (data) {
+    return ((data === false) ? '❌ API not available' : '✅ API available') + ' (' + JSON.stringify(data, null, 2) + ')';
```
englehardt, line 240: Nit: Since we consider these APIs harmful, shouldn't we flip the icons? Green=good, Red=Bad.

features/harmful-apis/main.js
```diff
@@ -0,0 +1,329 @@
...
+
+            if (result instanceof Promise) {
+                result
+                    .then(data => {
+                        valueSpan.innerHTML = resultToHTML(data);
+                        resultObj.value = data || null;
+                    })
+                    .catch(e => {
+                        failed++;
+                        valueSpan.innerHTML = `❌ error thrown ("${e.message ? e.message : e}")`;
```
englehardt, line 287: Nit: It would be great to error messages showed a ⚠️ to differentiate them from an expected failure

features/harmful-apis/main.js
```diff
@@ -0,0 +1,329 @@
...
+
+        all++;
+    }
+
+    updateSummary();
+
+    startButton.removeAttribute('disabled');
+}
+
+function downloadTheResults () {
```
englehardt, line 307: The downloaded results show `null` for false. Is that expected?

englehardt (changes requested):
A couple things need fixing, but overall this is in good shape. I was worried that a Secure context or the Permissions API may be required for some of these, but that doesn't seem to be the case for the properties I tested.

features/harmful-apis/main.js
```diff
@@ -0,0 +1,329 @@
...
+    {
+        id: 'floc-api',
+        run: () => {
+            return ('interestCohort' in document);
+        }
+    },
+    {
+        id: 'raw-sockets-api',
+        run: () => {
+            return ('openUDPSocket' in navigator) || ('openTCPSocket' in navigator);
```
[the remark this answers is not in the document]
kdzwinel: I added those opportunistically based on https://docs.google.com/document/d/1Xa5nFkIWxkL3hZHvDYWPhT8sZvNeFpCUKNuqIwZHxnE/edit# . No one shipped those yet - https://chromestatus.com/feature/6398297361088512 . Maybe we should note somehow that the test is based on a WIP spec?

features/harmful-apis/main.js
```diff
@@ -0,0 +1,329 @@
...
+    {
+        id: 'idle-detection-api',
+        run: () => {
+            return ('IdleDetector' in window);
+        }
+    },
+    {
+        id: 'raw-clipboard-api',
+        run: () => {
+            return ('clipboard' in navigator) && ('platform' in navigator.clipboard);
```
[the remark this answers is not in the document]
kdzwinel: The `platform` param didn't require prompt to go through (as far as I understand from the [spec](https://docs.google.com/document/d/1XDOtTv8DtwTi4GaszwRFIJCOuzAEA4g9Tk0HrasQAdE/edit)). However, it looks like the proposal is dead - https://chromestatus.com/feature/5682768497344512 ("no longer pursuing"), so perhaps we should remove it?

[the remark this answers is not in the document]
englehardt: Yes I'd say to remove it then.

features/harmful-apis/main.js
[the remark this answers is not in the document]
englehardt: Can you note in the output?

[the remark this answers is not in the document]
GrapeGreen: Randomly dropping by with a few things:

- This API is only supported in installed Chrome PWAs, and we currently have no plans on introducing it to the open web. 
- Yes, it also needs correct headers and a few other conditions to be fulfilled to evaluate to true :)
- The spec is severely outdated (same with the chrome status). I hope to devote some time in the coming weeks to reworking the github repo so that it better reflects the current state of the project.

That being said, I'm not sure that this test makes much sense atm :)

kdzwinel:
Note to self: add fledge -> https://developer.chrome.com/docs/privacy-sandbox/fledge/

features/harmful-apis/main.js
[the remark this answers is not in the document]
kdzwinel: Thanks both, I moved this API to manual testing, marked as something that hasn't shipped yet and noted the PWA requirement.

features/harmful-apis/main.js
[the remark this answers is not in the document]
kdzwinel: Removed!

features/harmful-apis/main.js
kdzwinel: Yeah, good call I changed it to thrown exception instead - this way the output looks correct. See screenshot in other comment.

features/harmful-apis/main.js
kdzwinel: I changed it so that there is only a red icon for APIs that are available. See screenshot in the other comment.

features/harmful-apis/main.js
kdzwinel: Only for failed tests, for other tests true/false. I fixed it.

features/harmful-apis/main.js
```diff
@@ -0,0 +1,329 @@
...
+        }
+    },
+    {
+        id: 'network-information-api',
+        run: () => {
+            return ('connection' in navigator);
+        }
+    },
+    {
+        id: 'client-hints-api',
```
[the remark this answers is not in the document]
kdzwinel: I think it's a good idea but it'd be nice to do this for all pages affected somehow. I'm putting it on a backlog for now.

kdzwinel:
Thanks for a review @englehardt ! All feedback should be addressed now, back to you.

harmful-apis

dharb (approved):
Played around with this a bit locally, everything looks to be in order. I'm going to go ahead and merge this now so that I can use it in my Windows privacy audit.

PR #79, opened 2022-03-24
Framed autofill

GioSensation:
Adds forms within iframes to test autofill.


autofill/frame-parent.htmk
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
```
GioSensation, line 1: @jonathanKingston This was in the branch already, but I suppose it can be deleted?

autofill/frame-parent.html
```diff
@@ -0,0 +1,30 @@
...
+
+<body>
+  <p><a href="../index.html">[Home]</a></p>
+
+  <p>We should show the autofill prompts for these iframed forms (note: it does <em>not</em> work on Android)</p>
+
+  <h2>Same origin iframe</h2>
+  <iframe src="./frame-child.html" id="frame-form"></iframe>
+  <h2>Cross origin iframe</h2>
+  <iframe src="https://good.third-party.site/autofill/frame-child.html" id="frame-form-cross-origin"></iframe>
```
GioSensation, line 19: I've added a cross-origin iframe for completeness as this was discussed extensively during the Android investigation, though it doesn't really make much difference for our purposes. I'm not entirely sure if using `https://good.third-party.site` is the correct way to do this, nor if it will actually work. AFAIK, we're not adding cross-origin iframes restrictions in CSPs so they should work by default. Please let me know if I should add anything else to make this work.

GioSensation:
@jonathanKingston I piggybacked your branch for https://app.asana.com/0/0/1201935429432145/f. Let me know what you think.

autofill/frame-parent.htmk
kdzwinel: Looks like something we should drop to me.

autofill/frame-parent.html
```diff
@@ -0,0 +1,30 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width">
+  <title>Form submission detection and autofill</title>
```
kdzwinel, line 7: let's update the title to match the name in index.html

autofill/frame-parent.html
kdzwinel: without knowing whole context this LGTM. good.third-party.site will work, path checks out. If you used other option (bad. , broken.) frame would be blocked/restricted.

kdzwinel (approved):
@GioSensation please update the title and drop that `htmk` file and feel free to merge 👍

jonathanKingston:
Yeah agree with Konrad's comments here, drop the htmk file :). Happy to make the changes if you're busy also.

PR #82, opened 2022-05-05
Bugfix: HSTS Test

englehardt:
This is a second bugfix for the HSTS test on the live site. I've actually verified this one manually this time, so hopefully it will be the last.

It seems we can't do the HSTS check on custom domains because Glitch doesn't provide a way to discover the original protocol for those domains. The header we're relying on to disclose the protocol of the client connection only works for the glitch domain.


kdzwinel (approved):
LGTM, thanks for verifying on glitch!

PR #83, opened 2022-05-19
Add demo phishing page

jonathanKingston:
Task: https://app.asana.com/0/0/1202305926765966/f

Not really much to it; just a test page we can use to block. I used a directory as we likely will want demo pages for other types of attacks.


kdzwinel (approved):
LGTM!

PR #86, opened 2022-07-12
update FB CTL test cases

ladamski:
Updates some FB CTL test cases so that share buttons, etc show up again

### DIFF
--- a/privacy-protections/click-to-load/index.html
+++ b/privacy-protections/click-to-load/index.html
@@ -6,33 +6,35 @@
   <title>Facebook click to load</title>
 </head>
 <body>
-  <!-- Load the facebook SDK in the recommended way -->
-  <script async defer crossorigin="anonymous" src="https://connect.facebook.net/en_US/sdk.js?XFBML=false"></script>
+<!-- Load the facebook SDK in the recommended way -->
+<div id="fb-root"></div>
+<script async defer crossorigin="anonymous" src="https://connect.facebook.net/en_US/sdk.js#xfbml=1&version=v14.0" nonce="1SgoaWbh"></script>
 <p><a href="../../">[Home]</a> ↣ <a href="../">[Privacy Protections Tests]</a> ↣ <strong>[Facebook click to load]</strong></p>
 
-<h2> About ClickToLoad Tests</h2>  
-<p>This page embeds the facebook SDK, and several facebook iFrames. It also attempts to create 
-a variety of social elements using the SDK.</P>
+<h2> About ClickToLoad Tests</h2>
+<p>This page embeds the facebook SDK, and several facebook iFrames. It also attempts to create
+a variety of social elements using the SDK.</p>
 
 <p>The page will count how many facebook requests have been sent. In some cases, the browser may start requesting iFrame links, which should be blocked by the extension (but will be counted in the Facebook calls). Calls should not exceed iFrames (and manual verification that those network calls were blocked is a good practice). Once elements are clicked, the count should increase and elements should display (in most cases - this page does not use a valid SDK APP ID, so not all functions work correctly, login in particular.).</p>
 
 <h2> Metrics </h2>
 <div id="facebook_calls"> Facebook Calls: <span id="facebook_call_count">0</span></div>
-<div id="facebook_calls"> Facebook iFrames: <span id="facebook_iFrames">0</span></div>
-<p><button id='download'>Download the result</button></p>
+<div id="facebook_iframes"> Facebook iFrames: <span id="facebook_iFrames">0</span></div>
+<p><button id="download">Download the result</button></p>
 
 <h2> All the social buttons from the SDK </h2>
-<div class="fb-like" data-href="https://developers.facebook.com/docs/plugins/1" data-width="" data-layout="standard" data-action="like" data-size="small" data-share="true"></div>
-<br /><div class="fb-like" data-href="https://developers.facebook.com/docs/plugins/2" data-width="" data-layout="standard" data-action="recommend" data-size="small" data-share="true"></div>
-<br /><div class="fb-save" data-uri="https://developers.facebook.com/docs/plugins/1" data-size="large"></div>
-<br /><div class="fb-share-button" data-href="https://developers.facebook.com/docs/plugins/" data-layout="button_count" data-size="small"><a target="_blank" href="https://www.facebook.com/sharer/sharer.php?u=https%3A%2F%2Fdevelopers.facebook.com%2Fdocs%2Fplugins%2F&amp;src=sdkpreparse" class="fb-xfbml-parse-ignore">Share</a></div>
+<div class="fb-like" data-href="https://privacy-test-pages.glitch.me/" data-width="" data-layout="standard" data-action="like" data-size="small" data-share="true"></div>
+<br><div class="fb-like" data-href="https://privacy-test-pages.glitch.me/" data-width="" data-layout="standard" data-action="recommend" data-size="small" data-share="true"></div>
+<br><div class="fb-save" data-uri="https://privacy-test-pages.glitch.me/" data-size="small"><p>Save button (appears to be broken in FB docs)</p></div>
+<br><div class="fb-share-button" data-href="https://privacy-test-pages.glitch.me/" data-layout="button_count" data-size="small"><a target="_blank" href="https://www.facebook.com/sharer/sharer.php?u=https%3A%2F%2Fdevelopers.facebook.com%2Fdocs%2Fplugins%2F&amp;src=sdkpreparse" class="fb-xfbml-parse-ignore">Share</a></div>
 
 <h2> All the social buttons in iFrames </h2>
-<iframe src="https://www.facebook.com/plugins/like.php?href=https%3A%2F%2Fdevelopers.facebook.com%2Fdocs%2Fplugins2%2F&width=450&layout=standard&action=like&size=small&share=true&height=35&appId" width="450" height="35" style="border:none;overflow:hidden" scrolling="no" frameborder="0" allowfullscreen="true" allow="autoplay; clipboard-write; encrypted-media; picture-in-picture; web-share"></iframe>
-<br /><iframe src="https://www.facebook.com/v2.0/plugins/like.php?href=https%3A%2F%2Fdevelopers.facebook.com%2Fdocs%2Fplugins1%2F&width=450&layout=standard&action=like&size=small&share=true&height=35&appId" width="450" height="35" style="border:none;overflow:hidden" scrolling="no" frameborder="0" allowfullscreen="true" allow="autoplay; clipboard-write; encrypted-media; picture-in-picture; web-share"></iframe>
-<br /><iframe src="https://www.facebook.com/plugins/share_button.php?href=https%3A%2F%2Fdevelopers.facebook.com%2Fdocs%2Fplugins%2F&layout=button_count&size=small&width=121&height=20&appId" width="121" height="20" style="border:none;overflow:hidden" scrolling="no" frameborder="0" allowfullscreen="true" allow="autoplay; clipboard-write; encrypted-media; picture-in-picture; web-share"></iframe>
+<iframe src="https://www.facebook.com/plugins/like.php?href=https%3A%2F%2Fprivacy-test-pages.glitch.me%2F&width=450&layout=standard&action=like&size=small&share=true&height=35&appId" width="450" height="35" style="border:none;overflow:hidden" scrolling="no" frameborder="0" allowfullscreen="true" allow="autoplay; clipboard-write; encrypted-media; picture-in-picture; web-share"></iframe>
+<br><iframe src="https://www.facebook.com/plugins/like.php?href=https%3A%2F%2Fprivacy-test-pages.glitch.me%2F&width=450&layout=standard&action=recommend&size=small&share=true&height=35&appId" width="450" height="35" style="border:none;overflow:hidden" scrolling="no" frameborder="0" allowfullscreen="true" allow="autoplay; clipboard-write; encrypted-media; picture-in-picture; web-share"></iframe>
+<p> SAVE button N/A in iframes per FB docs </p>
+<br><div class="fb-group" data-href="https://www.facebook.com/groups/ayearofrunning/" data-width="280" data-show-metadata="true"><blockquote cite="https://www.facebook.com/groups/ayearofrunning/" class="fb-xfbml-parse-ignore">A Year of Running</blockquote></div>
 
-<h2> Login buttons (SDK & custom) </h2>
+<h2> Login buttons (SDK &amp; custom) </h2>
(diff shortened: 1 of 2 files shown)


privacy-protections/click-to-load/index.html
```diff
@@ -6,33 +6,35 @@
...
-  <script async defer crossorigin="anonymous" src="https://connect.facebook.net/en_US/sdk.js?XFBML=false"></script>
+<!-- Load the facebook SDK in the recommended way -->
+<div id="fb-root"></div>
+<script async defer crossorigin="anonymous" src="https://connect.facebook.net/en_US/sdk.js#xfbml=1&version=v14.0" nonce="1SgoaWbh"></script>
 <p><a href="../../">[Home]</a> ↣ <a href="../">[Privacy Protections Tests]</a> ↣ <strong>[Facebook click to load]</strong></p>
 
-<h2> About ClickToLoad Tests</h2>  
-<p>This page embeds the facebook SDK, and several facebook iFrames. It also attempts to create 
-a variety of social elements using the SDK.</P>
+<h2> About ClickToLoad Tests</h2>
```
kzar, line 14: Nit: Could you make the whitespace inside of tags consistent? (Elsewhere too.) Personally I'd remove the extra spaces everywhere, but I don't mind much either way if it's consistent.

privacy-protections/click-to-load/index.html
```diff
@@ -6,33 +6,35 @@
...
+<iframe src="https://www.facebook.com/plugins/like.php?href=https%3A%2F%2Fprivacy-test-pages.glitch.me%2F&width=450&layout=standard&action=like&size=small&share=true&height=35&appId" width="450" height="35" style="border:none;overflow:hidden" scrolling="no" frameborder="0" allowfullscreen="true" allow="autoplay; clipboard-write; encrypted-media; picture-in-picture; web-share"></iframe>
+<br><iframe src="https://www.facebook.com/plugins/like.php?href=https%3A%2F%2Fprivacy-test-pages.glitch.me%2F&width=450&layout=standard&action=recommend&size=small&share=true&height=35&appId" width="450" height="35" style="border:none;overflow:hidden" scrolling="no" frameborder="0" allowfullscreen="true" allow="autoplay; clipboard-write; encrypted-media; picture-in-picture; web-share"></iframe>
+<p> SAVE button N/A in iframes per FB docs </p>
```
kzar, line 34: I don't understand this note about save buttons. Perhaps you could explain a bit more and link to the relevant part of the Facebook docs for context?

privacy-protections/click-to-load/index.html
```diff
@@ -6,33 +6,35 @@
...
+<!-- Load the facebook SDK in the recommended way -->
+<div id="fb-root"></div>
+<script async defer crossorigin="anonymous" src="https://connect.facebook.net/en_US/sdk.js#xfbml=1&version=v14.0" nonce="1SgoaWbh"></script>
 <p><a href="../../">[Home]</a> ↣ <a href="../">[Privacy Protections Tests]</a> ↣ <strong>[Facebook click to load]</strong></p>
 
-<h2> About ClickToLoad Tests</h2>  
-<p>This page embeds the facebook SDK, and several facebook iFrames. It also attempts to create 
-a variety of social elements using the SDK.</P>
+<h2> About ClickToLoad Tests</h2>
+<p>This page embeds the facebook SDK, and several facebook iFrames. It also attempts to create
```
kzar, line 15: Nit: It kind of bugs me how Facebook is capitalised inconsistently, would you mind making it consistent? (Elsewhere too.)

kzar (commented):
Changes look great @ladamski, I left a few suggestions.

privacy-protections/click-to-load/index.html
```diff
@@ -53,10 +55,10 @@ <h2> Objects in iFrames </h2>
...
 <br /><iframe src="https://www.facebook.com/plugins/post.php?href=https%3A%2F%2Fwww.facebook.com%2F20531316728%2Fposts%2F10154009990506729%2F&width=500&show_text=true&height=292&appId" width="500" height="292" style="border:none;overflow:hidden" scrolling="no" frameborder="0" allowfullscreen="true" allow="autoplay; clipboard-write; encrypted-media; picture-in-picture; web-share"></iframe>
 <br /><iframe src="https://www.facebook.com/plugins/video.php?href=https%3A%2F%2Fwww.facebook.com%2Ffacebook%2Fvideos%2F10153231379946729%2F&width=500&show_text=false&height=280&appId" width="500" height="280" style="border:none;overflow:hidden" scrolling="no" frameborder="0" allowfullscreen="true" allow="autoplay; clipboard-write; encrypted-media; picture-in-picture; web-share" allowFullScreen="true"></iframe>
-<p> Groups don't currently allow iFrame displays, but the documentation still shows this as a valid method of showing groups, so it's included here in case it changes again in the future </p> 
+<p> Groups don't currently allow iFrame displays, but the documentation still shows this as a valid method of showing groups, so it's included here in case it changes again in the future </p>
```
kzar, line 58: My vote would be to comment this one out too since it's expected not to work.

kdzwinel (commented):
I tried running locally and bunch of stuff didn't show + I got this error:

> Cannot listen to an undefined element. [Caught in: Tried to listen to element of type click]
> Subsequent non-fatal errors won't be logged; see https://fburl.com/debugjs.

This might be due to localhost/http though.

Since Dave started looking into this I'll leave the review to him 🙏

privacy-protections/click-to-load/index.html
ladamski: Yeah I didn't do a full lint because practically no line would have been left untouched, diff is hard enough to read as-is.  I'll do that through this pass.

privacy-protections/click-to-load/index.html
```diff
@@ -6,33 +6,35 @@
...
-<br /><div class="fb-like" data-href="https://developers.facebook.com/docs/plugins/2" data-width="" data-layout="standard" data-action="recommend" data-size="small" data-share="true"></div>
-<br /><div class="fb-save" data-uri="https://developers.facebook.com/docs/plugins/1" data-size="large"></div>
-<br /><div class="fb-share-button" data-href="https://developers.facebook.com/docs/plugins/" data-layout="button_count" data-size="small"><a target="_blank" href="https://www.facebook.com/sharer/sharer.php?u=https%3A%2F%2Fdevelopers.facebook.com%2Fdocs%2Fplugins%2F&amp;src=sdkpreparse" class="fb-xfbml-parse-ignore">Share</a></div>
+<div class="fb-like" data-href="https://privacy-test-pages.glitch.me/" data-width="" data-layout="standard" data-action="like" data-size="small" data-share="true"></div>
+<br><div class="fb-like" data-href="https://privacy-test-pages.glitch.me/" data-width="" data-layout="standard" data-action="recommend" data-size="small" data-share="true"></div>
+<br><div class="fb-save" data-uri="https://privacy-test-pages.glitch.me/" data-size="small"><p>Save button (appears to be broken in FB docs)</p></div>
```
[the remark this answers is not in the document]
ladamski: 👍

privacy-protections/click-to-load/index.html
ladamski: Iframes aren't supported for save button (which is being deprecated and I couldn't get to work anyway).  I just deleted this comment

privacy-protections/click-to-load/index.html
ladamski: 👍

privacy-protections/click-to-load/index.html
```diff
@@ -6,33 +6,35 @@
...
-<br /><iframe src="https://www.facebook.com/v2.0/plugins/like.php?href=https%3A%2F%2Fdevelopers.facebook.com%2Fdocs%2Fplugins1%2F&width=450&layout=standard&action=like&size=small&share=true&height=35&appId" width="450" height="35" style="border:none;overflow:hidden" scrolling="no" frameborder="0" allowfullscreen="true" allow="autoplay; clipboard-write; encrypted-media; picture-in-picture; web-share"></iframe>
-<br /><iframe src="https://www.facebook.com/plugins/share_button.php?href=https%3A%2F%2Fdevelopers.facebook.com%2Fdocs%2Fplugins%2F&layout=button_count&size=small&width=121&height=20&appId" width="121" height="20" style="border:none;overflow:hidden" scrolling="no" frameborder="0" allowfullscreen="true" allow="autoplay; clipboard-write; encrypted-media; picture-in-picture; web-share"></iframe>
```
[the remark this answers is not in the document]
ladamski: Good catch, copy/paste error, the "group" embed in the below comment was supposed to be "share"

privacy-protections/click-to-load/index.html
```diff
@@ -6,33 +6,35 @@
...
+<iframe src="https://www.facebook.com/plugins/like.php?href=https%3A%2F%2Fprivacy-test-pages.glitch.me%2F&width=450&layout=standard&action=like&size=small&share=true&height=35&appId" width="450" height="35" style="border:none;overflow:hidden" scrolling="no" frameborder="0" allowfullscreen="true" allow="autoplay; clipboard-write; encrypted-media; picture-in-picture; web-share"></iframe>
+<br><iframe src="https://www.facebook.com/plugins/like.php?href=https%3A%2F%2Fprivacy-test-pages.glitch.me%2F&width=450&layout=standard&action=recommend&size=small&share=true&height=35&appId" width="450" height="35" style="border:none;overflow:hidden" scrolling="no" frameborder="0" allowfullscreen="true" allow="autoplay; clipboard-write; encrypted-media; picture-in-picture; web-share"></iframe>
+<p> SAVE button N/A in iframes per FB docs </p>
+<br><div class="fb-group" data-href="https://www.facebook.com/groups/ayearofrunning/" data-width="280" data-show-metadata="true"><blockquote cite="https://www.facebook.com/groups/ayearofrunning/" class="fb-xfbml-parse-ignore">A Year of Running</blockquote></div>
```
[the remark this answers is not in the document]
ladamski: see above

privacy-protections/click-to-load/index.html
```diff
@@ -6,33 +6,35 @@
...
+<p>This page embeds the facebook SDK, and several facebook iFrames. It also attempts to create
+a variety of social elements using the SDK.</p>
 
 <p>The page will count how many facebook requests have been sent. In some cases, the browser may start requesting iFrame links, which should be blocked by the extension (but will be counted in the Facebook calls). Calls should not exceed iFrames (and manual verification that those network calls were blocked is a good practice). Once elements are clicked, the count should increase and elements should display (in most cases - this page does not use a valid SDK APP ID, so not all functions work correctly, login in particular.).</p>
 
 <h2> Metrics </h2>
 <div id="facebook_calls"> Facebook Calls: <span id="facebook_call_count">0</span></div>
-<div id="facebook_calls"> Facebook iFrames: <span id="facebook_iFrames">0</span></div>
-<p><button id='download'>Download the result</button></p>
+<div id="facebook_iframes"> Facebook iFrames: <span id="facebook_iFrames">0</span></div>
```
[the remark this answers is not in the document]
ladamski: 👍  yeah removed the div IDs

privacy-protections/click-to-load/index.html
ladamski: 👍

privacy-protections/click-to-load/index.html
```diff
@@ -6,33 +6,35 @@
...
+<p><button id="download">Download the result</button></p>
 
 <h2> All the social buttons from the SDK </h2>
```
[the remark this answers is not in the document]
ladamski: Not sure I understand which section you're referring to here?     The section below here?  I think "social" is correct, to be pedantic in FB-speak "share" and "like" aren't the same thing

ladamski:
Also a note on testing: the reason the buttons weren't showing up is because the href attributes weren't set correctly to the hosting website.  They don't allow "cross-domain liking" so you need to ensure the container host matches the refs.  So for https://concise-crocus-stetson.glitch.me/, I swapped "privacy-test-pages" with "concise-crocus-stetson"

privacy-protections/click-to-load/index.html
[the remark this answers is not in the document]
kzar: I guess I just mean that perhaps we can simplify this section of the test page. There's two sections that say "All social the buttons..." but then just have one button each. Perhaps we could combine the sections, or change the headings to "Share button..." for example?

image

But I don't insist if you think it's OK, just trying to think of ways we can make the page easier to understand. Also I think it is clearer now since you removed the broken example at least.

privacy-protections/click-to-load/index.html
```diff
@@ -1,62 +1,59 @@
...
+<script async defer crossorigin="anonymous" src="https://connect.facebook.net/en_US/sdk.js#xfbml=1&version=v14.0" nonce="1SgoaWbh"></script>
+<p><a href="../../">[Home]</a>↣<a href="../">[Privacy Protections Tests]</a>↣<strong>[Facebook click to load]</strong></p>
+
+<h2>About ClickToLoad Tests</h2>
+<p>This page embeds the Facebook SDK, and several Facebook iFrames. It also attempts to create
+a variety of social elements using the SDK.</p>
+
+<p>The page will count how many Facebook requests have been sent. In some cases, the browser may start requesting iFrame links, which should be blocked by the extension (but will be counted in the Facebook calls). Calls should not exceed iFrames (and manual verification that those network calls were blocked is a good practice). Once elements are clicked, the count should increase and elements should display (in most cases - this page does not use a valid SDK APP ID, so not all functions work correctly, login in particular.).</p>
+
+<h2>Metrics</h2>
```
kzar, line 20: @ladamski What do you think about:
1. removing the metrics section and the corresponding JS code
2. replacing the second stanza of the introduction with an explanation of what should happen when CTL is enabled (similar to how we do for [the YouTube CTL test page](https://privacy-test-pages.glitch.me/privacy-protections/youtube-click-to-load/)

My thinking is that the metrics aren't reliable, they seem effectively random and can't be trusted to verify the feature is working. They add noise to the page and complexity to the code, making both harder to understand. Also, they aren't required for the integration tests anyway.

I figure that having just a few notes to explain what the feature is expected to do (e.g. replace most FB with content, show a warning prompt when login buttons are clicked, etc) would be much more useful for folks trying to use the test page.

privacy-protections/click-to-load/index.html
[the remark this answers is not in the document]
ladamski: @kzar it should actually look like this (on https://concise-crocus-stetson.glitch.me/)
image

For FF at least I had to also disable the built-in privacy protections

privacy-protections/click-to-load/index.html
ladamski: Agree that is not super clear, the iframe count doesn't really depend on protection status anyway.  The "Facebook calls" one though should be zero when protections are enabled, so I think it should be replaced with a simpler boolean metric that remains "BLOCKED" unless a call is detected.

I like the idea of adding an intro paragraph, that would be very useful!

kzar: This is a huge improvement, nice one! :1st_place_medal:

kzar: Optional: Perhaps rename "Facebook Resource Loads" to something like "Automatic Facebook Resource Loads"? Doesn't really roll off the tongue, but I'm thinking that it might be kind of misleading to keep saying "NONE" otherwise after the user clicked to unblock some Facebook content?

PR #87, opened 2022-07-13
update loads dynamically after initial load

ladamski:
Update FB load stats dynamically after initial window load, when unblocking elements, etc.

### DIFF
--- a/privacy-protections/click-to-load/main.js
+++ b/privacy-protections/click-to-load/main.js
@@ -1,5 +1,5 @@
 let facebookCalls = 0;
-let loadsUpdated = false;
+let pageLoaded = false;
 
 // object that contains results of all tests
 const results = {
@@ -65,8 +65,8 @@ function facebookObserver (list, observer) {
     }
 
     updateResults();
-    if (loadsUpdated) {
-        updateLoads();
+    if (pageLoaded || facebookCalls) {
+        displayLoads();
     }
 }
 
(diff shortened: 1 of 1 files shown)


privacy-protections/click-to-load/main.js
```diff
@@ -86,6 +90,11 @@ function downloadTheResults () {
 const downloadButton = document.querySelector('#download');
 downloadButton.addEventListener('click', () => downloadTheResults());
 
-window.onload = function () {
+const updateLoads = function () {
```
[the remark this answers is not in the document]
kzar: Maybe call this `displayResult`?

privacy-protections/click-to-load/main.js
```diff
@@ -64,6 +65,9 @@ function facebookObserver (list, observer) {
     }
 
     updateResults();
+    if (loadsUpdated) {
```
kzar, line 68: Just an idea, feel free to ignore. But we could put this check inside of `updateLoads`. That way, we could display "DETECTED" right away when a request was detected (even before they all finish), but still wait until the end to display "NONE".

privacy-protections/click-to-load/main.js
```diff
@@ -86,6 +90,11 @@ function downloadTheResults () {
 const downloadButton = document.querySelector('#download');
 downloadButton.addEventListener('click', () => downloadTheResults());
 
-window.onload = function () {
+const updateLoads = function () {
     document.getElementById('facebook_call_count').innerHTML = facebookCalls ? '<span style="color:RED;">DETECTED</span>' : '<span style="color:GREEN;">NONE</span>';
```
kzar, line 94: I think using `innerHTML` is kind of frowned upon. How about having the span element there already, but with an ID, and then setting the span's class/style + text here?

kzar (commented):
This is great. I think this change will make the page much easier to understand :+1:.

privacy-protections/click-to-load/main.js
```diff
@@ -86,6 +90,11 @@ function downloadTheResults () {
...
 };
+
+window.onload = function () {
+    updateLoads();
+    loadsUpdated = true;
+};
```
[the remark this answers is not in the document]
ladamski: This is all thanks to the linter setup, will be red otherwise.

privacy-protections/click-to-load/main.js
ladamski: innerHTML is crude but effective, no issue in this case, its not like we have a bunch of event listeners hanging off that.  I'd rather KISS

privacy-protections/click-to-load/main.js
ladamski: Tweaked

PR #89, opened 2022-07-25
Add a bunch of test cases for autofill autoprompt

GioSensation:
**Asana:** https://app.asana.com/0/0/1202658859281356/f
**Reviewer:** @shakyShane 

# Description
Adds a bunch of test cases for autofill autoprompt. The expected behaviour is added to the pages themselves.

### DIFF
--- a/autofill/autoprompt/case-1.html
+++ b/autofill/autoprompt/case-1.html
@@ -1,0 +1,54 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width">
+  <title>Autofill autoprompt for signin forms</title>
+  <link rel="stylesheet" href="../style.css" />
+</head>
+
+<body>
+<p><a href="../index.html">[Home]</a></p>
+<nav>
+  <ul class="nav">
+    <li><a href="./case-1.html">Case 1</a></li>
+    <li><a href="./case-2.html">Case 2</a></li>
+    <li><a href="./case-3.html">Case 3</a></li>
+    <li><a href="./case-4.html">Case 4</a></li>
+    <li><a href="./case-5.html">Case 5</a></li>
+  </ul>
+</nav>
+
+<p id="demo"></p>
+
+<div class="">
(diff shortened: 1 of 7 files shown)


shakyShane (approved):
🚀

autofill/autoprompt/case-1.html
```diff
@@ -0,0 +1,54 @@
...
+
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width">
+  <title>Autofill autoprompt for signin forms</title>
+  <link rel="stylesheet" href="../style.css" />
+</head>
+
+<body>
+<p><a href="../index.html">[Home]</a></p>
```
kdzwinel, line 12: this doesn't work because we are two levels deep (autofill/autoprompt/). Same for all test case pages.

index.html
```diff
@@ -75,6 +75,16 @@ <h2>Autofill</h2>
 	<li><a href="./autofill/modal.html">Email autofill form within self-closing modal</a></li>
 	<li><a href="./autofill/form-submission.html">Form submission detection and autofill</a></li>
 	<li><a href="./autofill/frame-form-submission-parent.html">Form submission detection and autofill within an iframe</a></li>
+  <li>
+		Autoprompt
```
kdzwinel, line 79: Can we expand a bit more what autoprompt is? I'm guessing something like "automatically prompt to store credentials after login form is detected"?

index.html
```diff
@@ -75,6 +75,16 @@ <h2>Autofill</h2>
...
+		<ul>
+			<li><a href="./autofill/autoprompt/case-1.html">Case 1</a></li>
```
kdzwinel, line 81: It'd be good for those cases to explain them a bit, so that we don't have to rely on remembering the numbers e.g. "Form below the fold" instead of "Case 3".

kdzwinel (commented):
Sorry, I'm late to the party. I left couple of small comments.

GioSensation:
Thanks for the feedback, @kdzwinel. Very valid points. I've fixed them in https://github.com/duckduckgo/privacy-test-pages/pull/91.

PR #90, opened 2022-07-29
add a page for testing autofill for login forms embedded in an iframe

szanto90balazs:
Asana: https://app.asana.com/0/72649045549333/1201816478631739/f

Add a new page for testing autofill for logins within an iframe. There are related pages ([parent](https://github.com/duckduckgo/privacy-test-pages/blob/main/autofill/frame-parent.html), [child](https://github.com/duckduckgo/privacy-test-pages/blob/main/autofill/frame-child.html)) to test autofill with iframes, but they target the autofill email integration and not logins. Felt like adding a new page makes more sense instead of extending the already existing one to avoid breaking tests or whatnot.

### DIFF
--- a/autofill/frame-form-submission-child.html
+++ b/autofill/frame-form-submission-child.html
@@ -1,0 +1,41 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width">
+  <title>Form submission detection and autofill within an iframe</title>
+  <link rel="stylesheet" href="./style.css" />
+</head>
+
+<body>
+  <p id="demo"></p>
+
+  <div class="">
+    <p>Submitting this form should prompt to save credentials.</p>
+
+    <form id="click-form" novalidate action="/autofill/form-submission.html">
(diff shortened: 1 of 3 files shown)


autofill/frame-form-submission-child.html
```diff
@@ -0,0 +1,41 @@
...
+      </fieldset>
+    </form>
+    <script>
+      const clickForm = document.getElementById('click-form');
+      const clickFormButton = clickForm.querySelector('button');
+      clickFormButton?.addEventListener('click', (e) => {
+          e.preventDefault();
+          e.stopImmediatePropagation();
+          // Note that form.submit() does not fire a submit event
+          clickForm.submit();
```
GioSensation, line 34: I suggest also adding a visual indication that the form was indeed submitted. See [this](https://github.com/duckduckgo/privacy-test-pages/pull/89/files#diff-864a4e1a3d54def77f5ee9e7507d27bc22cf8fe3472ae053d8ce6184b3f78059R46) as an example.

GioSensation (approved):
Just a minor suggestion, but it's approved. 👍

PR #93, opened 2022-08-17
Ad attribution test flow

jonathanKingston:
Sample test cases for working through testing our platforms click attribution code.


index.html
```diff
@@ -22,6 +22,12 @@ <h2>Tracker Reporting</h2>
 	<li><a href="./tracker-reporting/1major-via-fetch.html">1 major tracker loaded via fetch</a> <a href="./tracker-reporting/1major-via-fetch.html?delay=5000">[Delay 5s]</a></li>
 </ul>
 
+<h2>Ad flow</h2>
+
+<ul>
+	<li><a href="./adClickFlow/serp/index.html">Ad flow</a></li>
+</ul>
```
kdzwinel, line 29: This doesn't really explain what this test does. We don't provide any additional info in `serp/index.html` either. I think we should leave at least a bit of explanation.

adClickFlow/shared/testCases.json
```diff
@@ -0,0 +1,494 @@
...
+            },
+            {
+              "url": "https://www.ad-company.site/ping.gif",
+              "status": "parent blocked"
+            }
+          ]
+        }
+      }
+    ]
+  },
```
englehardt, line 349: You could expand these test cases with Ads 9 - 14 in https://app.asana.com/0/72649045549333/1202521973600829/f

server.js
```diff
@@ -49,6 +49,76 @@ app.use(json({
 const partitioningRoutes = require('./privacy-protections/storage-partitioning/server/routes');
 app.use('/partitioning', partitioningRoutes);
 
+function joinPath () {
```
englehardt, line 52: Is it possible to move these into the adClickFlow routes.js?

englehardt (approved):
LGTM module some suggestions.

server.js
jonathanKingston: Not really, these rewrite the path to the relevant _/adClickFlow_ path. The /adClickFlow/server/routes.js is applied only to that path with _app.use('/adClickFlow', adClickFlowRoutes);_


The only other way here is export a function from the server path; I'd like to punt this to a fast follow commit that has a separate approval as we've already been testing this well.

adClickFlow/shared/testCases.json
jonathanKingston: Yeah that's the plan, I just wanted us to merge the current code then I can start on these.

README.md
```diff
@@ -26,6 +26,11 @@ We have couple of test domains, that all resolve to `privacy-test-pages.glitch.m
 - `good.third-party.site` - non-tracking third party, it's not on our blocklist and will not be blocked by our clients
 - `broken.third-party.site` - tracking third party that we can't block (e.g. due to brekage), it's on our blocklist, but it will not be blocked by our clients
 - `bad.third-party.site` - tracking third party that's on our blocklist and our clients will block
+- `www.search-company.example` - Simulated search provider
+- `convert.ad-company.example` - Simulated ad provider conversion ping
+- `www.ad-company.example` - Simulated ad provider non-conversion ping
+- `www.payment-company.example` - Simulated payment provider
+- `www.publisher-company.example`- Simulated publisher website
```
kdzwinel, line 33: > that all resolve to `privacy-test-pages.glitch.me`

Sorry, just caught that - these were all supposed to be `.site` not `.example`, right?

PR #96, opened 2022-08-29
Enable single test runs on referrer trimming

SlayterDev:
This PR passes around the testid as well as adds manual test links to run single tests on this page.


kdzwinel (commented):
I understand why we want to have those shortcuts in the HTML, but why do we need to pass this ID around is unclear to me? We get testid from the url param, we pass it to the test, it ends up in the url to the come-back page, gets passed back by the come-back page in the url - and then I'm not sure what happens?

SlayterDev:
When the test page runs the test it:
- Checks if the test results are stored locally and adds them to the page
- checks if the results for the last test are in the url and puts them in the results section in the page
- otherwise runs the test

It does this for each test in the array. Currently if we pass a testid when we run the test we get the filtered array which calls the specified test. When the page comes back with the results we've lost the id so it operates on the full array of tests which puts the last run in the wrong spot and continues with the others.

With this change passing along the testid, when we come back to the results page the testid will ensure the array of tests is filtered again and will stop with just that one test instead of continuing with the others.

kdzwinel (approved):
Gotcha, that makes sense, thanks for explanation!

kdzwinel:
Sorry, I had to revert it because it was breaking the main functionality of "start test". The way I tested (Safari):
- I've run couple of manual tests
- I've tried running the "Start test" (with storage cleared and clean url)

Expected: full test runs normally
Actual: only one test runs and no results are displayed

PR #98, opened 2022-08-31
Enable single test runs

SlayterDev:
Same as last PR but fixes "run tests" issue.


kdzwinel (approved):
Thanks for fixing! Let's try again.

PR #104, opened 2022-11-02
Client hints test page.

kdzwinel:
Lists hint values from headers and JS.

Example from testing in Edge:
edge-ch

Doesn't explode in browsers that don't support CH (!!).

### DIFF
--- a/features/client-hints/main.html
+++ b/features/client-hints/main.html
@@ -10,12 +10,13 @@
 <body>
   <p><a href="../index.html">[Home]</a></p>
 
-  <p>This site asks browser for all client hints possible via headers and via JS.</P>
+  <p>This site asks browser for all client hints possible via headers and via JS.</p>
+  <p>To test a default set of client hints sent go <a href='https://good.third-party.site/features/client-hints/'>here</a> - this loads the same test, but doesn't send the <code>accept-ch</code> header.</P>
 
-  <h3>Hints requested via `accept-ch`:</h3>
+  <h3>Hints received via request headers:</h3>
   <table id="headers"></table>
 
-  <h3>Hints requested via `navigator.userAgentData`:</h3>
+  <h3>Hints requested via <code>navigator.userAgentData</code>:</h3>
   <table id="js"></table>
 
   <script src="./all-ch.js"></script>
--- a/features/client-hints/main.html
+++ b/features/client-hints/main.html
@@ -1,0 +1,97 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width">
+  <title>Client Hints</title>
+</head>
+
+<body>
+  <p><a href="../index.html">[Home]</a></p>
+
+  <p>This site asks browser for all client hints possible via headers and via JS.</P>
+
+  <h3>Hints requested via `accept-ch`:</h3>
+  <table id="headers"></table>
+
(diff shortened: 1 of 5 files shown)


features/client-hints/main.html
```diff
@@ -0,0 +1,97 @@
...
+
+    fetch('/reflect-headers')
+        .then(r => r.json()).then(data => {
+            ALL_CH.forEach(hint => {
+                const value = data.headers && data.headers[hint];
+
+                headerTable.appendChild(createRow(hint, value));
+            });
+
+            Object.keys(data.headers).forEach(header => {
```
kdzwinel, line 65: Besides checking for hints that we requested, we are also looking for any hints that we haven't requested and flagging them.

features/client-hints/main.html
```diff
@@ -0,0 +1,97 @@
...
+
+    if (navigator.userAgentData) {
+        navigator.userAgentData.getHighEntropyValues(ALL_CH_JS).then(hints => {
+            ALL_CH_JS.forEach(hint => {
+                const value = Array.isArray(hints[hint]) ? JSON.stringify(hints[hint]) : hints[hint];
+
+                jsTable.appendChild(createRow(hint, value));
+            });
+
+            Object.keys(hints).forEach(hint => {
```
kdzwinel, line 82: ditto just in case, looking for hints that we haven't requested

server.js
```diff
@@ -333,3 +334,15 @@ app.get('/features/download/:type', (req, res) => {
         break;
     }
 });
+
+const ALL_CH_STR = ALL_CH.join(',');
+app.get('/features/client-hints/', (req, res) => {
+    res.set('accept-ch', ALL_CH_STR);
```
kdzwinel, line 340: this could have been a meta tag, but I wanted to have one source of truth for CH list (all-ch.js). This also has a cool side effect that if you go to http://localhost:3000/features/client-hints/main.html directly header will NOT be sent showing you what's sent by default

englehardt: > directly header will NOT be sent showing you what's sent by default

Only if you haven't already visited a site that has client hints sent? At least that's what it looked like when I played around with it.

It might be nice to have a page where we never request full CH so we can easily test the default. Maybe make use of one of the extra domains for that (e.g., `www.first-party.site`)?

server.js
```diff
@@ -333,3 +334,15 @@ app.get('/features/download/:type', (req, res) => {
         break;
     }
 });
+
+const ALL_CH_STR = ALL_CH.join(',');
+app.get('/features/client-hints/', (req, res) => {
```
englehardt, line 339: It would be nice to place this routes in the client hints directory, rather than continuing to expand the big server.js file. See how I did it for storage partitioning here :https://github.com/duckduckgo/privacy-test-pages/blob/main/privacy-protections/storage-partitioning/server/routes.js and here: https://github.com/duckduckgo/privacy-test-pages/blob/main/server.js#L48-L50

englehardt (approved):
LGTM with some minor suggestions.

server.js
kdzwinel: Good call, I'll add an expeception for one domain.

server.js
kdzwinel: Yep, 100%, in fact I'll do a follow up PR to clean up some of the old routes.

kdzwinel:
Thanks for taking a look @englehardt , I implemented both suggestions.

PR #105, opened 2022-11-04
Expand tracker blocking test (redirect and iframe^2)

kdzwinel:
Adds two test for edge-cases that we encountered over the last few months:
- tracker loaded via a safe redirect (-> good 307 -> bad)
- tracker loaded from an iframe within an iframe


jonathanKingston (approved):
These look great, sorry for the delay in review!

PR #106, opened 2022-11-04
Show cookie expiry if cookieStore available.

kdzwinel:
expiry

Limitations:
- only JS set cookies (on header set cookies we have httponly flag)
- only Chromium (CookieStore API)

### DIFF
--- a/privacy-protections/storage-blocking/3rdparty.js
+++ b/privacy-protections/storage-blocking/3rdparty.js
@@ -7,14 +7,22 @@
 (function () {
     const src = document.currentScript.src;
     const trackingDomain = src.indexOf('https://broken.third-party.site/') === 0;
+    const cookieName = trackingDomain ? 'tptdata' : 'tpsdata';
 
     commonTests.push({
         id: `JS cookie (3rd party ${trackingDomain ? 'tracking' : 'safe'} script)`,
         store: (data) => {
-            document.cookie = `tp${trackingDomain ? 't' : 's'}data=${data}; expires= Wed, 21 Aug 2030 20:00:00 UTC;`;
+            document.cookie = `${cookieName}=${data}; expires= Wed, 21 Aug 2030 20:00:00 UTC;`;
         },
         retrive: () => {
             return trackingDomain ? document.cookie.match(/tptdata=([0-9]+)/)[1] : document.cookie.match(/tpsdata=([0-9]+)/)[1];
+        },
+        extra: () => {
+            if (window.cookieStore) {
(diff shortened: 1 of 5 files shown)


privacy-protections/storage-blocking/3rdparty.js
```diff
@@ -7,14 +7,22 @@
...
     commonTests.push({
         id: `JS cookie (3rd party ${trackingDomain ? 'tracking' : 'safe'} script)`,
         store: (data) => {
-            document.cookie = `tp${trackingDomain ? 't' : 's'}data=${data}; expires= Wed, 21 Aug 2030 20:00:00 UTC;`;
+            document.cookie = `${cookieName}=${data}; expires= Wed, 21 Aug 2030 20:00:00 UTC;`;
         },
         retrive: () => {
             return trackingDomain ? document.cookie.match(/tptdata=([0-9]+)/)[1] : document.cookie.match(/tpsdata=([0-9]+)/)[1];
+        },
+        extra: () => {
```
kdzwinel, line 20: I introduced an "extra" callback that can show additional info next to the result. This info doesn't go into the results object.

privacy-protections/storage-blocking/iframe.js
```diff
@@ -31,33 +31,28 @@ function storeData (randomNumber) {
 }
 
 function retrieveData () {
-    return Promise.all(commonTests.map(test => {
+    return Promise.all(commonTests.map(async (test) => {
         try {
             const result = test.retrive();
+            const value = (result instanceof Promise) ? await result : result;
```
sammacbeth, line 37: Await also works on non-promises.
```suggestion
            const value = await result;
```

privacy-protections/storage-blocking/iframe.js
```diff
@@ -31,33 +31,28 @@ function storeData (randomNumber) {
...
-                        error: e.message
-                    }));
-            } else {
-                return Promise.resolve({
-                    test: test.id,
-                    value: result
-                });
+            if (test.extra) {
+                const extraResult = test.extra();
+                extra = (extraResult instanceof Promise) ? await extraResult : extraResult;
```
sammacbeth, line 42:
```suggestion
                extra = await extraResult;
```
As above

sammacbeth (approved):
Looks good. One minor comment, but non-blocking.

privacy-protections/storage-blocking/iframe.js
kdzwinel: TIL, this is great. Thanks!

PR #108, opened 2022-11-07
Add Card and Address form page for autofill

Bunn:
Add two new autofill form pages: Card and Address


kdzwinel (approved):
LGTM cc @GioSensation

kdzwinel:
Thanks Fernando! Merging… should be on production in 2min.

PR #109, opened 2022-11-15
Run cookie header tests in frames.

sammacbeth:
Moves the storage header cookie sites to the list of common tests, so that they also run in safe and tracker iframes.

### DIFF
--- a/privacy-protections/storage-blocking/helpers/commonTests.js
+++ b/privacy-protections/storage-blocking/helpers/commonTests.js
@@ -1,53 +1,35 @@
 /* exported commonTests */
 /* global cookieStore, THIRD_PARTY_TRACKER_ORIGIN, THIRD_PARTY_ORIGIN */
 
+function generateCookieHeaderTest (namePrefix, origin) {
+    return {
+        id: `${namePrefix} header cookie`,
+        store: (data) => {
+            return fetch(`${origin}/set-cookie?value=${data}`, { credentials: 'include' }).then(r => {
+                if (!r.ok) {
+                    throw new Error('Request failed.');
+                }
+            });
+        },
+        retrive: () => {
+            return fetch(`${origin}/reflect-headers`, { credentials: 'include' })
+                .then(r => r.json())
+                .then(data => data.headers.cookie.match(/headerdata=([0-9]+)/)[1]);
+        }
+    };
+}
+
+const cookieHeaderTests = [generateCookieHeaderTest('first party', '')];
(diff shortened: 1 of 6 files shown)


privacy-protections/storage-blocking/main.js
```diff
@@ -278,7 +231,7 @@ function retrieveData () {
         });
     }
 
-    tests.concat(commonTests).forEach(test => {
```
kdzwinel, line 281: fancy!

privacy-protections/storage-blocking/helpers/commonTests.js
```diff
@@ -1,7 +1,53 @@
...
+            });
+        },
+        retrive: () => {
+            return fetch('/reflect-headers')
+                .then(r => r.json())
+                .then(data => data.headers.cookie.match(/headerdata=([0-9]+)/)[1]);
+        }
+    },
+    {
+        id: 'safe third party header cookie',
```
kdzwinel, line 22: hm, it's a bit awkward that in iframes this and next test are redundant (`${THIRD_PARTY_ORIGIN}/set-cookie` === `/set-cookie`). They are also not third party in this context so the description is wrong. Can we somehow elegantly exclude them from running in frames where origins match?

privacy-protections/storage-blocking/helpers/commonTests.js
```diff
@@ -1,7 +1,35 @@
...
+            return fetch(`${origin}/set-cookie?value=${data}`, { credentials: 'include' }).then(r => {
+                if (!r.ok) {
+                    throw new Error('Request failed.');
+                }
+            });
+        },
+        retrive: () => {
+            return fetch(`${origin}/reflect-headers`, { credentials: 'include' })
+                .then(r => r.json())
+                .then(data => data.headers.cookie.match(/headerdata=([0-9]+)/)[1]);
```
kdzwinel, line 17: One more thing, all of those tests are setting a `headerdata` cookie on the same set of two domains (THIRD_PARTY_TRACKER_ORIGIN, THIRD_PARTY_ORIGIN). Some of them will be set in a first party context, some of them in a third party context. So it may happen that a cookie will be set in a first-party context but be reported in a third party context or vice-versa. Should we prefix cookie names somehow to avoid confusion? e.g. 'firstparty-headerdata' and 'thirdparty-headerdata'?

kdzwinel:
Thanks a lot for those changes! Figuring out those cookies won't be straightforward, but it does make sense when you think about it for a bit. Merging!

oh-boy

kdzwinel (approved):

PR #111, opened 2023-01-18
Element hiding test page

dharb:
This is a test page for the element hiding feature. I've set it up so that it will work if it is served on either the `first-party.site` or `privacy-test-pages.glitch.me` domains (see relevant rules in privacy config [here](https://github.com/duckduckgo/privacy-configuration/blob/ac318bebf711087b7254559574e65989bfe85329/features/element-hiding.json#L648-L690)).

To run locally (on macOS) I had to add entries to my `/etc/hosts` file pointing `good.third-party.site` & `www.first-party.site` -> `127.0.0.1`, and then followed the mkcert instructions from the README, but with `third-party.site "*.third-party.site"` & `first-party.site "*.first-party.site"` appended to the command--this actually produced `first-party.example+13-key.pm` and `first-party.example+13-key.pm` files, which weren't picked up by express without editing [these lines](https://github.com/duckduckgo/privacy-test-pages/blob/main/server.js#L29-L40) to match. @kdzwinel should I update these and the README docs to match?

The page tests several cases for each rule type and outputs the results to a table as well as a `results` global variable.

cc @jonathanKingston 

image


features/element-hiding/index.html
```diff
@@ -0,0 +1,121 @@
...
+<body>
+    <p><a href="../../">[Home]</a> ↣ [Features] ↣ <strong>[Element Hiding Test Page]</strong></p>
+
+    <p>Test page for element hiding, a feature used to hide DOM elements. Tests take a few seconds to run.</p>
+
+    <table id="results-table">
+        <tr><th>Description</th><th>Outcome</th></tr>
+    </table>
+
+    <div class="spacer">
```
kdzwinel, line 48: nit, but instead of this you could just say that below will be debug data like I did in https://privacy-test-pages.glitch.me/privacy-protections/request-blocking/ - makes debugging slightly easier if you have to load the page multiple times

features/element-hiding/main.js
```diff
@@ -0,0 +1,128 @@
+const TEST_DOMAIN = 'good.third-party.site';
```
kdzwinel, line 1: nice job extracting this to the top, makes testing locally easier

features/element-hiding/main.js
```diff
@@ -0,0 +1,128 @@
...
+    { id: 'closest-empty-single-nested-div', description: 'closest-empty rule: parent of target element is hidden', hidden: true },
+    { id: 'closest-empty-siblings-nested-div', description: 'closest-empty rule: parent of target element with empty siblings is hidden', hidden: true },
+    { id: 'closest-empty-siblings-content-nested-div', description: 'closest-empty rule: parent of target element with non-empty siblings is not hidden', hidden: false },
+    { id: 'closest-empty-frame-content', description: 'closest-empty rule: parent of element containing iframe is not hidden', hidden: false },
+    { id: 'closest-empty-delayed-frame', description: 'closest-empty rule: parent of element that loads 3p frame after delay is not hidden', hidden: false },
+    { id: 'closest-empty-intermediate-element', description: 'closest-empty rule: only outermost element should be hidden when target is nested', hidden: false },
+    { id: 'closest-empty-delayed-parent', description: 'closest-empty rule: parent of empty element added to page after delay is hidden', hidden: true },
```
kdzwinel, line 20: not sure if that's the test, but are we testing the "unhiding" after content is loaded into a frame that we thought is empty?

features/element-hiding/main.js
```diff
@@ -0,0 +1,128 @@
...
+    } else {
+        testCell.innerText = 'Failed';
+        testCell.style.backgroundColor = '#f97268';
+    }
+
+    results.results.push(result);
+}
+
+function injectDelayedContent () {
+    setTimeout(function () {
```
kdzwinel, line 83:
```suggestion
    setTimeout(() => {
```

(we use arrow function mostly everywhere now)

features/element-hiding/main.js
```diff
@@ -0,0 +1,128 @@
...
+
+    const descriptionCell = row.insertCell(0);
+    const testCell = row.insertCell(1);
+
+    // set default values and colors
+    descriptionCell.innerText = description;
+
+    const result = {
+        id: name,
+        hiddenProperly: testPassed
```
kdzwinel, line 68: some tests are testing if element is visible making `hiddenProperly` not the best name :) Can be `result` as in other places (I realize the awkwardness of `result['x'].result`)

kdzwinel (commented):
Looks and works great, nicely done! I left some comments, mostly nits.

features/element-hiding/main.js
dharb: So this one is actually testing that the parent element (in this case, the most distant ancestor that has no visible content) is hidden after an element we target with a `closest-empty-test` class is added to it.

Let me add a test for the case when a frame we thought was empty has content added to it.

features/element-hiding/main.js
dharb: Old habits die hard 😢

dharb:
@kdzwinel thanks for the review! Just pushed a commit to address your comments.

kdzwinel:
LGTM, thank you!

PR #113, opened 2023-02-03
Add info for autofill features

GioSensation:
Just adds a couple of disclaimers to avoid false alarms when testing autofill across different platforms.

### DIFF
--- a/autofill/frame-parent.html
+++ b/autofill/frame-parent.html
@@ -12,6 +12,7 @@
   <p><a href="../index.html">[Home]</a></p>
 
   <p>We should show the autofill prompts for these iframed forms (note: it does <em>not</em> work on Android)</p>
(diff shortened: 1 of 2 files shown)


autofill/frame-parent.html
```diff
@@ -20,7 +21,7 @@ <h2>Cross origin iframe</h2>
 
   <style>
     iframe {
-      height: 2em;
+      height: 4em;
```
GioSensation, line 24: This ensures that a) the tooltip sort of shows (it's not completely hidden) and b) it shows that it's cut off in extensions but not on macOS.

kzar: Good idea :+1:

kzar (approved):
LGTM

PR #114, opened 2023-02-06
Fix a typo in fingerprinting tests

muodov:


kdzwinel (approved):
Thanks! I wonder if this will break any integration tests out there 😬  We will see I guess!

PR #118, opened 2023-02-20
Add support for Safari 16.1

jonathanKingston:
https://app.asana.com/0/547792610048271/1203993845372678/f

### DIFF
--- a/security/js-leaks.html
+++ b/security/js-leaks.html
@@ -29,6 +29,7 @@
         <option value="chrome_110">Chrome 110</option>
         <option value="safari_14">Safari 14</option>
         <option value="safari_14_ios">Safari iOS 14</option>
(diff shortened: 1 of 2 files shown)


security/js-leaks.html
```diff
@@ -29,6 +29,7 @@
         <option value="chrome_110">Chrome 110</option>
         <option value="safari_14">Safari 14</option>
         <option value="safari_14_ios">Safari iOS 14</option>
+        <option value="safari_16_1">Safari 16.1</option>
```
kdzwinel, line 32: Nit - I think we should only keep the latest version and remove the old ones (safari 14). I don't think there is any use for them?

kdzwinel (approved):
Thanks!

PR #128, opened 2023-03-09
test page for API attribution 

muodov:
---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204144475536788

### DIFF
--- a/crawler/attribution/entrypoints/blob-url.js
+++ b/crawler/attribution/entrypoints/blob-url.js
@@ -1,5 +1,8 @@
 {
-    const b = new Blob([`console.log("${document.currentScript.src}", navigator.userAgent); window.addResult('Navigator.prototype.userAgent', 'script in a blob url', '${document.currentScript.src}');`], { type: 'text/javascript' });
+    const b = new Blob([`
+    console.log("${document.currentScript.src}", navigator.userAgent);
+    window.addResult('Navigator.prototype.userAgent', 'script in a blob url', location.href, '${document.currentScript.src}');
+    `], { type: 'text/javascript' });
     const u = URL.createObjectURL(b);
     const s = document.createElement('script');
     s.src = u;
--- a/crawler/attribution/entrypoints/blob-url.js
+++ b/crawler/attribution/entrypoints/blob-url.js
@@ -1,0 +1,8 @@
+{
+    const b = new Blob([`console.log("${document.currentScript.src}", navigator.userAgent)`], { type: 'text/javascript' });
+    const u = URL.createObjectURL(b);
+    const s = document.createElement('script');
+    s.src = u;
+    document.body.appendChild(s);
+    window.addResult('Navigator.prototype.userAgent', 'script in a blob url', document.currentScript.src);
+}
--- a/crawler/attribution/entrypoints/blob-url.js
+++ b/crawler/attribution/entrypoints/blob-url.js
@@ -1,8 +1,7 @@
 {
-    const b = new Blob([`console.log("${document.currentScript.src}", navigator.userAgent)`], { type: 'text/javascript' });
+    const b = new Blob([`console.log("${document.currentScript.src}", navigator.userAgent); window.addResult('Navigator.prototype.userAgent', 'script in a blob url', '${document.currentScript.src}');`], { type: 'text/javascript' });
     const u = URL.createObjectURL(b);
     const s = document.createElement('script');
     s.src = u;
     document.body.appendChild(s);
-    window.addResult('Navigator.prototype.userAgent', 'script in a blob url', document.currentScript.src);
 }
--- a/crawler/attribution/entrypoints/createelement.js
+++ b/crawler/attribution/entrypoints/createelement.js
@@ -1,6 +1,8 @@
 {
     const el = document.createElement('script');
-    el.textContent = `console.log("${document.currentScript.src}", navigator.userAgent)`;
+    el.textContent = `
+    console.log("${document.currentScript.src}", navigator.userAgent);
+    window.addResult('Navigator.prototype.userAgent', 'inline script via document.createElement()', location.href, '${document.currentScript.src}');
+    `;
     document.body.appendChild(el);
-    window.addResult('Navigator.prototype.userAgent', 'inline script via document.createElement()', document.currentScript.src);
 }
--- a/crawler/attribution/entrypoints/createelement.js
+++ b/crawler/attribution/entrypoints/createelement.js
@@ -1,0 +1,6 @@
+{
+    const el = document.createElement('script');
+    el.textContent = `console.log("${document.currentScript.src}", navigator.userAgent)`;
+    document.body.appendChild(el);
+    window.addResult('Navigator.prototype.userAgent', 'inline script via document.createElement()', document.currentScript.src);
+}
--- a/crawler/attribution/entrypoints/data-url.js
+++ b/crawler/attribution/entrypoints/data-url.js
@@ -1,7 +1,6 @@
 {
-    const u = `data:application/javascript,console.log("${document.currentScript.src}", navigator.userAgent)`;
+    const u = `data:application/javascript,console.log("${document.currentScript.src}", navigator.userAgent);window.addResult('Navigator.prototype.userAgent', 'script in a data: url', '${document.currentScript.src}')`;
     const s = document.createElement('script');
     s.src = u;
     document.body.appendChild(s);
-    window.addResult('Navigator.prototype.userAgent', 'script in a data: url', document.currentScript.src);
 }
--- a/crawler/attribution/entrypoints/data-url.js
+++ b/crawler/attribution/entrypoints/data-url.js
@@ -1,5 +1,5 @@
 {
-    const u = `data:application/javascript,console.log("${document.currentScript.src}", navigator.userAgent);window.addResult('Navigator.prototype.userAgent', 'script in a data: url', '${document.currentScript.src}')`;
+    const u = `data:application/javascript,console.log("${document.currentScript.src}", navigator.userAgent);window.addResult('Navigator.prototype.userAgent', 'script in a data: url', location.href, '${document.currentScript.src}')`;
     const s = document.createElement('script');
     s.src = u;
     document.body.appendChild(s);
--- a/crawler/attribution/entrypoints/data-url.js
+++ b/crawler/attribution/entrypoints/data-url.js
@@ -1,0 +1,7 @@
+{
+    const u = `data:application/javascript,console.log("${document.currentScript.src}", navigator.userAgent)`;
+    const s = document.createElement('script');
+    s.src = u;
+    document.body.appendChild(s);
+    window.addResult('Navigator.prototype.userAgent', 'script in a data: url', document.currentScript.src);
+}
--- a/crawler/attribution/entrypoints/deep-async-stack.js
+++ b/crawler/attribution/entrypoints/deep-async-stack.js
@@ -3,7 +3,7 @@ async function deepCallAsync () {
         await deepCallAsync();
     } else {
         console.log(deepCallAsync.src, navigator.userAgent);
-        window.addResult('Navigator.prototype.userAgent', 'deep async call stack', deepCallAsync.src);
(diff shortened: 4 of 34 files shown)


crawler/attribution/entrypoints/blob-url.js
```diff
@@ -0,0 +1,8 @@
+{
```
kdzwinel, line 1: why do you have brackets on all those files? To prevent const/let leaking to the global context?

crawler/attribution/entrypoints/prototype-overload.js
```diff
@@ -0,0 +1,21 @@
...
+    );
+    const origImpl = descriptor.value;
+    const src = document.currentScript.src;
+    descriptor.value = function () {
+        console.log(src, navigator.userAgent);
+        window.addResult('Navigator.prototype.userAgent', 'createElement prototype overload', src);
+        return origImpl.apply(this, arguments);
+    };
+
+    Object.defineProperty(
```
kdzwinel, line 14: doesn't this influence other tests that use createElement?

crawler/attribution/index.html
```diff
@@ -0,0 +1,89 @@
...
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width">
+  <title>API call attribution test</title>
+</head>
+
+<body>
+<p><a href="../../index.html">[Home]</a> ↣ <strong>[API call attribution test]</strong></p>
+
+<p>This page tests various ways of script injection to verify that API calls are attributed to the correct script URL.</p>
```
kdzwinel, line 11: I think it's worth noting that the API being tested is `navigator.userAgent`.

crawler/attribution/entrypoints/deep-async-stack.js
```diff
@@ -0,0 +1,13 @@
+{
+    const src = `${document.currentScript.src}`;
+    let depth = 0;
+    async function deepCall () {
+        if (depth++ < 1000) {
```
kdzwinel, line 5: nit but in webkit for some reason this and next step throw errors that "depth" doesn't exist 🤷

kdzwinel (commented):
This is great, looks super thorough! I left couple of small comments. BTW for some reason this page never loads for me - possibly not a big deal but if we wanted to automate testing of this page waiting for load might be an useful signal that won't work here.

crawler/attribution/entrypoints/blob-url.js
jonathanKingston, line 1: Not that I really care, that's the justification for the { in these files?

crawler/attribution/entrypoints/data-url.js
```diff
@@ -0,0 +1,7 @@
+{
+    const u = `data:application/javascript,console.log("${document.currentScript.src}", navigator.userAgent)`;
```
jonathanKingston, line 2: The result here isn't being used, shouldn't we be calling addResult from inside data?

crawler/attribution/entrypoints/blob-url.js
```diff
@@ -0,0 +1,8 @@
...
+    const b = new Blob([`console.log("${document.currentScript.src}", navigator.userAgent)`], { type: 'text/javascript' });
```
jonathanKingston, line 2: I think we should be addResult-ing here no?

crawler/attribution/entrypoints/blob-url.js
muodov: yep, exactly.

crawler/attribution/entrypoints/prototype-overload.js
kdzwinel: We could potentially run this test in an iframe to make sure it doesn't have side-effects, but not sure if worth it. Up to you.

crawler/attribution/sw-source.js
```diff
@@ -0,0 +1,15 @@
...
+    self.skipWaiting();
+});
+
+self.addEventListener('activate', (evt) => {
+    evt.waitUntil(self.clients.claim());
+});
+
+self.addEventListener('message', (event) => {
+    console.log(self.location.href, navigator.userAgent);
+    event.source.postMessage({ dave: self.location.href });
```
jonathanKingston, line 13: Can we name this something meaningful? (or are we doing this elsewhere?)

crawler/attribution/entrypoints/blob-url.js
jonathanKingston: This only happens with var right?

crawler/attribution/iframe-sandbox.html
```diff
@@ -0,0 +1,3 @@
+<script>
+    console.log(location.href, navigator.userAgent);
```
jonathanKingston, line 2: Should this be postmessaging it's URL?

crawler/attribution/entrypoints/prototype-overload.js
muodov: I thought about it, but right now running things in an iframe makes it more flaky

crawler/attribution/entrypoints/blob-url.js
muodov: In this case top-level `const` variables do collide (i'm getting an exception). I think these scripts share the same scope or something like that...

crawler/attribution/entrypoints/data-url.js
muodov: Good catch, thanks!

crawler/attribution/entrypoints/blob-url.js
muodov: Yup, you're right. Changed it

crawler/attribution/sw-source.js
muodov: done, sorry :)

crawler/attribution/entrypoints/createelement.js
```diff
@@ -0,0 +1,6 @@
+{
+    const el = document.createElement('script');
+    el.textContent = `console.log("${document.currentScript.src}", navigator.userAgent)`;
```
jonathanKingston, line 3: addResult from inside again?

crawler/attribution/entrypoints/dom0.js
```diff
@@ -0,0 +1,5 @@
+{
+    const el = document.querySelector('#container');
+    el.innerHTML = `<img src=/ onerror="console.log('${document.currentScript.src}', navigator.userAgent)" />`;
```
jonathanKingston, line 3: Add result from the onerror?

crawler/attribution/entrypoints/eval.js
```diff
@@ -0,0 +1,6 @@
+function doEval () {
+    // eslint-disable-next-line no-eval
+    eval(`console.log("${document.currentScript.src}", navigator.userAgent)`);
```
jonathanKingston, line 3: addResult here from eval?

crawler/attribution/entrypoints/eventlistener.js
```diff
@@ -0,0 +1,9 @@
+{
+    const img = document.createElement('img');
+    img.src = '/';
+    const src = `${document.currentScript.src}`;
+    img.addEventListener('error', () => {
+        console.log(src, navigator.userAgent);
```
jonathanKingston, line 6: Add result from inside the listener?

Is a regular function any different also?

crawler/attribution/entrypoints/iframe-data-url.js
```diff
@@ -0,0 +1,7 @@
+{
+    const u = `data:text/html,<script>console.log("${document.currentScript.src}", navigator.userAgent);</script>`;
```
jonathanKingston, line 2: addResult from inside?

crawler/attribution/entrypoints/iframe-sandbox.js
```diff
@@ -0,0 +1,7 @@
+{
+    const f = document.createElement('iframe');
+    f.src = 'iframe-sandbox.html';
+    f.sandbox = 'allow-scripts';
+    document.body.appendChild(f);
+    window.addResult('Navigator.prototype.userAgent', 'inline script inside sandboxed iframe', f.src);
```
jonathanKingston, line 6: addResult from inside the sandbox or message pass back out?

crawler/attribution/entrypoints/prototype-overload.js
```diff
@@ -0,0 +1,21 @@
...
+        Document.prototype,
+        'createElement',
+        descriptor
+    );
+
+    document.createElement('div');
```
jonathanKingston, line 20: This doesn't capture the intent of what I meant here. Another script would be calling this probably 1P.

crawler/attribution/entrypoints/eval.js
jonathanKingston, line 3: Should we test `new Function('....')` also?

crawler/attribution/iframe-sandbox.html
muodov: good idea!

crawler/attribution/entrypoints/createelement.js
muodov: I've reworked addResult() across all places

crawler/attribution/entrypoints/prototype-overload.js
muodov: gotcha, I've added another case and moved most of the JS entry points to 3p

crawler/attribution/entrypoints/eval.js
muodov: good idea, I've added a test

crawler/attribution/entrypoints/prototype-overload.js
jonathanKingston: This is the 1p hooked the method and the 3p called it; which is interesting (similar to the jQuery case we discussed).

I was hoping we'd cater for the other way around where 3p hooks the method and then 1p calls it.

crawler/attribution/iframe-sandbox.html
jonathanKingston: Should we be printing out the result and sending it through 'messageFromSandbox'?

jonathanKingston: Ignore sorry we're not printing the result elsewhere. (I think we should but it can be a follow up, as I think we should allow for flexibility of the API we're triggering also)

muodov: Initially I wanted to pass the API name to the test cases, but it would require some text preprocessing or eval. There's no real use for it right now because this test is about the attribution algorithm, and that is supposed to not depend on specific API

crawler/attribution/entrypoints/eventlistener.js
jonathanKingston: I just mean engine wise they're quite different so might diverge in implementation.

Happy to ignore also.

crawler/attribution/entrypoints/web-worker.js
```diff
@@ -0,0 +1,4 @@
+const worker = new Worker('worker-source.js');
+worker.addEventListener('message', function (event) {
+    window.addResult('Navigator.prototype.userAgent', 'web worker', event.data.workerUrl, event.data.workerUrl);
```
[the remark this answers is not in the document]
muodov: The call is inside worker-source.js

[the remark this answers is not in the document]
jonathanKingston: Ignore, this is the message passing from the worker.

It's a little easy to forget the callsite we're checking and where we are sending the result from. Maybe add a comment for this and service worker?

crawler/attribution/entrypoints/eventlistener.js
muodov: Right now it doesn't make a difference on the crawling results. If it does in the future we will need to cover both cases, but for now I think we can leave it as is

crawler/attribution/entrypoints/deep-async-stack.js
muodov: Good catch! There's some weirdness in the way closures work in webkit, I made a workaround

PR #129, opened 2023-03-10
AdClick tests to validate non-302 redirects

RendijsSmukulis:
- Added two adClick tests to help validate ad attribution works when 301 and 307 redirects are used as well.
- Both the tests are based on the `[Ad 5]` test, but have different redirects used.


jonathanKingston:
Sorry missed this, thanks!

jonathanKingston (approved):

PR #132, opened 2023-03-21
Shared content scope page tests also used for integration tests

jonathanKingston:
Requires: https://github.com/duckduckgo/content-scope-scripts/pull/329 to move.

- Updated CI version as it doesn't like external http packages (also doesn't match nvm)
- Brings in dependabot to automate these test changes.
- Tests are moved to content-scope-scripts just to reduce changes of adding more.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "ws": "^7.4.0"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


kdzwinel (approved):
LGTM, let's see how it goes!

PR #138, opened 2023-03-22
Install dependencies after deploy

kdzwinel:
We now have C-S-S as an prod dep.

https://app.asana.com/0/1186013049913869/1204236919542000/f

### DIFF
--- a/git.sh
+++ b/git.sh
@@ -8,3 +8,6 @@ git reset --hard origin/main
 
 # Force pull
 git pull origin main --force
+
+# Install dependencies
+npm i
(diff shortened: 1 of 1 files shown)


git.sh
```diff
@@ -8,3 +8,6 @@ git reset --hard origin/main
 
 # Force pull
 git pull origin main --force
+
+# Install dependencies
+npm i
```
jonathanKingston, line 13:
```suggestion
npm ci
```

This won't try and update anything and will install only the package-lock.json checkins.

Otherwise 👍🏻  I hope this works!

PR #150, opened 2023-04-11
Ignore scripts on install, move C-S-S to be a build dep

jonathanKingston:


jonathanKingston:
Merging as is a blocker to verification.

PR #161, opened 2023-05-16
Remove 404'd AMP link

SlayterDev:
The euronews.com link now returns a 404 which negates its value in this test.


kdzwinel:
LGTM, thanks!

PR #165, opened 2023-09-21
Migrate app off glitch.me onto own hosted infra

KeaganJarvis:
With this app now running on our own infra, this change replaces references to the old third party host and their domain name/URL. And now points to the current, self hosted domain/URL for this app. 

Asana: https://app.asana.com/0/0/1205442219109576/f


README.md
```diff
@@ -21,14 +21,14 @@ Please note that we are not taking external contributions for new test pages, bu
...
 
-We have couple of test domains, that all resolve to `privacy-test-pages.glitch.me`, which help us simulate various scenarios:
+We have couple of test domains, that all resolve to `privacy-test-pages.site`, which help us simulate various scenarios:
 
 - `www.first-party.site` - an alternative first-party domain used for tests that require first-party resources on other subdomains (e.g., `hsts.first-party.site`)
 - `good.third-party.site` - non-tracking third party, it's not on our blocklist and will not be blocked by our clients
 - `broken.third-party.site` - tracking third party that we can't block (e.g. due to brekage), it's on our blocklist, but it will not be blocked by our clients
 - `bad.third-party.site` - tracking third party that's on our blocklist and our clients will block
 
-We also have additional test domains that resolve to `ad-attribution-test-pages.glitch.me`, specifically for the Ad Attribution tests hosted [here](https://www.search-company.site):
```
kdzwinel, line 31: this is not right, this domain doesn't exist (ad-attribution-test-pages.site). We only had ad-attribution-test-pages.glitch.me because of glitch limitation, so perhaps we should reword it to:

> We also have additional test domains specifically for the Ad Attribution tests hosted [here] …

privacy-protections/storage-partitioning/main.js
```diff
@@ -252,8 +252,8 @@ if (window.location.hostname === 'localhost') {
     runButton.setAttribute('disabled', '');
 }
 
-// warn the user if loaded over privacy-test-pages.glitch.me
-if (window.location.hostname === 'privacy-test-pages.glitch.me') {
```
kdzwinel, line 256: some of those changes mean that `privacy-test-pages.glitch.me` will no longer work fully as expected - here it's not a big deal but just FYI

privacy-protections/storage-partitioning/server/routes.js
```diff
@@ -93,7 +93,7 @@ router.get('/get_hsts.png', (req, res) => {
     // The X-Forwarded-Proto header is added by Glitch's proxy
```
kdzwinel, line 93: hm, do we have that header? Probably not, right? Because it was glitches' weirdness? If so, we need to update this code to work on our server and correctly distinguish http from https - I can look into that if you are not familiar with node. Again, if we make this change for privacy-test-pages.sites, it means that glitch.me version will no longer funciton correctly, but I think in this case it's easy to support both if needed

server.js
```diff
@@ -232,7 +232,7 @@ app.get('/come-back', (req, res) => {
     const jsReferrer = document.referrer;
     document.body.innerHTML += '<p>header: <strong>${req.headers.referer || ''}</strong></p><p>js: <strong>' + jsReferrer + '</strong></p>';
     setTimeout(() => {
-        location.href = 'https://privacy-test-pages.glitch.me/privacy-protections/referrer-trimming/?run&header=${req.headers.referer || ''}&js=' + jsReferrer + '&testid=${req.query.testid || ''}';
+        location.href = 'https://privacy-test-pages.site/privacy-protections/referrer-trimming/?run&header=${req.headers.referer || ''}&js=' + jsReferrer + '&testid=${req.query.testid || ''}';
```
kdzwinel, line 235: This will basically redirect some of the tests happening on glitch.me to privacy-test-pages.site mid test. Might not be a big deal if we don't care about glitch.me anymore

kdzwinel (commented):
Thanks Keagan and sorry for a slow turn-around. I left some comments - happy to discuss sync.

privacy-protections/storage-partitioning/main.js
kdzwinel: Closing - we agreed that we will decommission glitch.me

server.js
kdzwinel: Closing - we agreed that we will decommission glitch.me

README.md
KeaganJarvis: Updated

privacy-protections/storage-partitioning/server/routes.js
KeaganJarvis: I'm not too familiar with the code discrepancies that are required to correctly work between http and https here.

kdzwinel: Per our discussion - this flag is set correctly above (`let isHTTPS = req.protocol === 'https';`) so we should be able to safely remove the whole if block below together with the comment. Let's test -> remove -> ship it

privacy-protections/storage-partitioning/server/routes.js
```diff
@@ -90,13 +90,6 @@ router.get('/set_hsts.png', (req, res) => {
 
 router.get('/get_hsts.png', (req, res) => {
     let isHTTPS = req.protocol === 'https';
-    // The X-Forwarded-Proto header is added by Glitch's proxy
-    // and reveals the original protocol used during the connection
-    // This header will always show HTTPS for all custom domains,
-    // it's only correct for privacy-test-pages.glitch.me.
-    if (req.headers['x-forwarded-proto']) {
-        isHTTPS = req.headers['x-forwarded-proto'].split(',', 1)[0] === 'https';
-    }
```
KeaganJarvis, line 99: @kdzwinel i ran a location.protocol and got back "https" in the console.

kdzwinel (approved):
Thanks a lot Keagan! This LGTM. Not sure if I can merge right away, so I'm leaving that to you.

PR #166, opened 2023-09-29
Add storage test for setting a cookie via the cookieStore API in a serviceworker

sammacbeth:
- As the `cookieStore` API is available in serviceworkers, this can circumvent protections that rely on content-scripts.

### DIFF
--- a/privacy-protections/storage-blocking/helpers/commonTests.js
+++ b/privacy-protections/storage-blocking/helpers/commonTests.js
@@ -158,7 +158,7 @@ const commonTests = [
         },
         extra: () => {
             if (window.cookieStore) {
-                return cookieStore.get('cookiestoredata').then(cookie => {
+                return cookieStore.get('swcookiestoredata').then(cookie => {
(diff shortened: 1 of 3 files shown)


privacy-protections/storage-blocking/helpers/commonTests.js
```diff
@@ -158,7 +158,7 @@ const commonTests = [
         },
         extra: () => {
             if (window.cookieStore) {
-                return cookieStore.get('cookiestoredata').then(cookie => {
+                return cookieStore.get('swcookiestoredata').then(cookie => {
```
kdzwinel, line 161: I'm confused why we have swcookiestoredata here and cookiestoredata the file below? Is this intended? They don't match up with their store/retrive functions now.

kdzwinel:
@KeaganJarvis  are we OK making new changes to this project? Or should we wait with merging as you are still setting things up?

privacy-protections/storage-blocking/helpers/commonTests.js
sammacbeth: Sorry, I got the cookie names mixed up. Now I've swapped them around.

kdzwinel:
I'm assuming we are good to merge.

PR #170, opened 2023-11-03
Add test for surrogate included in head tag

sammacbeth:
Adds a new surrogate test to check that scripts loaded in the `head` have the surrogate correctly loaded.

This test fails on iOS and macOS. Passes on other platforms.
You can try it out at https://test.smbco.de/privacy-protections/surrogates/


kdzwinel (approved):
Looks good, works great, thanks!

PR #178, opened 2023-12-01
Improve usability of the surrogate script test page

kzar:
The surrogate script test page was a little hard to use. Some expected
results being shown as red, some unexpected results being shown
green. Let's improve that a bit, making it clear to folks what a
success/failure looks like.


kdzwinel (approved):
Thanks Dave, this LGTM! Just a note that I think this might be a first time that we are indicating "pass" and "fail" directly, but I think it's warranted - this page was causing a lot of confusion.

PR #183, opened 2024-01-24
Autofill shadow dom test page

GioSensation:
Adds a test page with a form within a shadow tree. Also makes navigations between autofill pages easier by linking directly to the autofill section of index.html.


GioSensation:
Self merging.

PR #185, opened 2024-02-02
Windows Browser: Script Injection test pages

RendijsSmukulis:
### Description

This adds a couple of (for now) Window Browser specific test pages that allows testing script injection (e.g. content scope scripts):

- Validate injection in the main page
- Validate injection in an iframe
- Validate injection into a 3rd party iframe

It validates:
- Whether the scripts were injected
- Whether posting a message from native code to injected JS works
- Whether JS can send a message to native and receive a response
- Whether scripts are isolated from page scripts

Note that this requires running a modified version of Windows Browser that's intended for testing purposes, so the tests will not work with a release version.


muodov (approved):
lgtm, just one typo suggestion

PR #192, opened 2024-04-05
Add website that auto-refreshes every 3 seconds

bwaresiak:
Asana: https://app.asana.com/0/0/1207008140035284/f

Simple website to test if URL bar doesn't lose keyboard focus on refresh.

### DIFF
--- a/features/auto-refresh.html
+++ b/features/auto-refresh.html
@@ -8,7 +8,8 @@
 <body>
 <p><a href="../index.html">[Home]</a></p>
 
-<h1>Website that refreshes every 3 seconds. You can use this to validate URL bar focus.</h1>
+<h1>Website that refreshes every 3 seconds.</h1>
+<p>You can use this to validate that the URL bar keeps focus between reloads.</p>
 
 <script>
     setTimeout(function () {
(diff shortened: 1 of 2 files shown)


bwaresiak:
@kdzwinel / @jonathanKingston can one of you review this? thanks!

features/auto-refresh.html
```diff
@@ -0,0 +1,20 @@
...
+<html>
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width">
+  <title>Auto Refresh</title>
+</head>
+<body>
+<p><a href="../index.html">[Home]</a></p>
+
+<h1>Website that refreshes every 3 seconds. You can use this to validate URL bar focus.</h1>
```
kdzwinel, line 11:
```suggestion
<h1>Website that refreshes every 3 seconds.</h1>
<p>You can use this to validate that the URL bar keeps focus between reloads.</p>
```

I didn't first get it from the description, so I propose we expand it a bit.

kdzwinel (approved):
One small description comment, but otherwise LGTM. Thanks!

PR #199, opened 2024-04-23
Re-add content-scope-scripts

jonathanKingston:
We have a bunch of pages that should immediately become accessible after this change. (eg: https://github.com/duckduckgo/content-scope-scripts/blob/main/integration-test/test-pages/webcompat/index.html would be served from: https://privacy-test-pages.site/content-scope-scripts/web-compat/index.html) 

It's about time we re-add this back to allow this path.


kdzwinel (approved):
LGTM

PR #213, opened 2024-06-27
create separate signup form for ui testing

brindy:
**Background**
Adding UI tests, the current example sign up form immediately prompts to sign in which makes it different to test.

**Change**
Add a separate sign up page to the auto prompt and updated related files.


amddg44 (approved):
LGTM!

PR #215, opened 2024-07-08
add cookie setting page

brindy:
Adds a page that allows the user to increment a cookie by pressing a button and shows known text values for use in UI testing.


brindy:
We've got a page already

PR #219, opened 2024-08-01
Add expected pixels to ad click attribution test cases

kzar:
"Pixel" requests are sent by the mechanism DuckDuckGo relies on to
support ad click attribution in DuckDuckGo products. These help
DuckDuckGo validate that the mechanism's logic is working correctly.

Let's note the expected pixels in these shared test cases, so that
each DuckDuckGo product can ensure those requests are being made
correctly.

Further reading:
 - https://duckduckgo.com/duckduckgo-help-pages/privacy/web-tracking-protections/#duckduckgo-private-search-ads
 - https://improving.duckduckgo.com/

### DIFF
--- a/adClickFlow/shared/testCases.json
+++ b/adClickFlow/shared/testCases.json
@@ -9,7 +9,8 @@
           "url": "https://www.search-company.site/#ad-id-1"
         },
         "expected": {
-          "url": "https://www.search-company.site/#ad-id-1"
+          "url": "https://www.search-company.site/#ad-id-1",
+          "pixels": []
         }
       },
       {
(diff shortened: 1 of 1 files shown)


adClickFlow/shared/testCases.json
```diff
@@ -211,6 +220,29 @@
...
+              }
+            },
+            {
+              "name": "m_ad_click_active",
+              "params": {
+                "appVersion": "APP_VERSION"
+              }
+            },
+            {
+              "name": "m_pageloads_with_ad_attribution",
```
englehardt, line 241: There is an assumption here that for the test runner that we force this to fire at the end of each test case. I think that's okay, but just checking that it's expected.

englehardt (approved):

adClickFlow/shared/testCases.json
kzar: Yea, that's right. I would have mentioned that in a comment (like I did in the other version of this file) but unfortunately JSON doesn't support comments.

PR #241, opened 2024-11-12
Remove vox AMP test

sammacbeth:
Link is now a 404


kdzwinel (approved):
Thanks Sam, we need to migrate those from external to internal tests.

PR #249, opened 2024-12-17
Malicious site protection edge cases

kdzwinel:
I bundled couple of small changes:
- renamed "phishing detection" to "malicious site protection" to cover both phishing and malware and follow latest naming of this feature
- reorganized order of malicious site protection test pages on the index page for easier reading
- added three edge-case's: when site is in the local cache and doesn't require API request, when site is flagged by both DNS protection and malicious site protection, when site is excluded from protection via remote config exception

### DIFF
--- a/security/badware/index.html
+++ b/security/badware/index.html
@@ -3,27 +3,30 @@
 
 <head>
     <meta charset="UTF-8">
-    <title>Test Pages - Phishing Detection</title>
+    <title>Test Pages - Malicious Site Protection</title>
 </head>
 
 <body>
-    <h1>Phishing Detection Test Pages</h1>
+    <h1>Malicious Site Protection Test Pages</h1>
     <a href="/">[Home]</a>
     <ul>
         <li><a href="./phishing.html">Standard Phishing Test</a></li>
         <li><a href="./malware.html">Standard Malware Test</a></li>
-        <li><a href="./malware-download.html">Malware Download Test</a></li>
         <li><a href="./phishing-iframe-loader.html">Phishing iFrame Loader</a></li>
-        <li><a href="./phishing-js-redirector-helper.html">Phishing JS Redirector (Direct)</a></li>
-        <li><a href="./phishing-js-redirector.html">Phishing JS Redirector (Indirect)</a></li>
         <li><a href="./phishing-legit-iframe-loader.html">Phishing Legit iFrame Loader</a></li>
-        <li><a href="./phishing-meta-redirect-clean.html">Phishing Redirect via Meta Refresh (Not Flagged in Dataset)</a></li>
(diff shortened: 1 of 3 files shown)


security/badware/index.html
```diff
@@ -3,27 +3,30 @@
 
 <head>
```
kdzwinel, line 4: The page looks like this after changes:

render

security/badware/index.html
```diff
@@ -34,6 +37,18 @@ <h1>Phishing Detection Test Pages</h1>
...
+    <h2>Edge Cases</h2>
+    <ul>
+        <li><a href="http://malware.privacy-test-pages.site/security/badware/malware.html">Page flagged by both DNS protection and in-browser protection</a> (⚠️ different domain)</li>
+        <li><a href="http://broken.privacy-test-pages.site/security/badware/malware.html">Page where malicious site protection is disabled via remote config exception</a> (⚠️ different domain)</li>
+        <li><a href="./local-check.html">Page flagged in the local data (doesn't require API request)</a></li>
+    </ul>
+
+    <h2>Utility</h2>
+    <ul>
+        <li><a href="./phishing-iframe-top-navigator.html">Phishing iFrame Top Navigator</a></li>
```
kdzwinel, line 50: BTW, do we even have to link to it? I think it only confuses folks since it's not a test.

security/badware/index.html
```diff
@@ -3,27 +3,30 @@
...
-        <li><a href="./phishing-meta-redirect.html">Phishing Redirect via Meta Refresh (Flagged in Dataset)</a></li>
         <li><a href="./phishing-popups.html">Phishing Open via Popups</a></li>
         <li><a href="./phishing-url-tampering.html">Phishing Opening with URL Tampering</a></li>
         <li><a href="./phishing-form-submission.html">Phishing Form Submission</a></li>
-        <li><a href="./phishing-iframe-top-navigator.html">Phishing iFrame Top Navigator</a></li>
         <li><a href="./phishing-service-worker.html">Phishing Service Worker</a></li>
+        <li><a href="./malware-download.html">Malware Download Test</a></li>
+    </ul>
+
+    <h2>Redirects</h2>
```
kdzwinel, line 24: Not sure if it's a useful grouping, just wanted to make things more findable given we now have a ton of those pages. I'm happy to revert this change if it doesn't make sense

not-a-rootkit: No this makes perfect sense. Thanks Konrad!

security/badware/index.html
not-a-rootkit: We can remove this link actually, it's not useful on its own.

not-a-rootkit (approved):
Happy with this, appreciate it @kdzwinel !

security/badware/index.html
kdzwinel: Awesome, I dropped it

PR #261, opened 2025-03-03
Add uncommon URL escape sequence test page

not-a-rootkit:
We discovered that certain escape sequences may break our protections. After discussing with Netcraft, they confirmed that they make no attempt to URL decode before building regular expressions. Therefore, all URL encoding should be taken as is by the clients.

This PR introduces one such phishing sample page.

### DIFF
--- a/security/badware/index.html
+++ b/security/badware/index.html
@@ -40,6 +40,7 @@
 
     <h2>Edge Cases</h2>
     <ul>
+        <li><a href="./phishing badͮ.html">Phishing Page With Abnormal URL Escape Sequences</a></li>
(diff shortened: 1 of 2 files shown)


security/badware/index.html
```diff
@@ -40,6 +40,7 @@ <h2>Redirects</h2>
 
     <h2>Edge Cases</h2>
     <ul>
+        <li><a href="./phishing badͮ.html">Phishing Page With Abnormal URL Escape Sequences</a></li>
```
kdzwinel, line 43: I see that cheeky "dͮ". Is that the only trick here? Or is that space a part of it?

PR #319, opened 2025-08-29
JS-leaks: Add recent Chrome profile + expand collection to non-enumerable props

laghee:
**Asana:** https://app.asana.com/1/137249556945/task/1211021483002732?focus=true

- Added a current version of Chrome.
- Previously we weren't catching our shimmed Notification API because we were only grabbing enumerable properties.
- We catch a couple hundred more properties by both enumerating and using `Object.getOwnPropertyNames`

👉  `Object.getOwnPropertyNames` only:
Screenshot 2025-08-29 at 16 00 03

👉  `Object.getOwnPropertyNames` + enumeration:
Screenshot 2025-08-29 at 16 15 33


kdzwinel (commented):
Left one comment, but otherwise LGTM. Sounds like @jonathanKingston had something to add though? I don't see his comment.

jonathanKingston (approved):
Nice hadn't seen this PR. Thanks for clearing up the changes from the other one!

